PR #1, opened 2023-04-27
added functionality for delete_word_left, delete_word_right and move_left/right_by_word for text_input

aang521:


aang521:
Hi, that explains a lot 😅
I will take a look at it tomorrow. Would you prefer it it text input is also treated as byte offsets, or rather have me change my changes to treat it as character offsets?

focus-editor:
I think byte offsets is the way to go. I haven't touched that code in a long time, but I think there may be a couple of auxiliary functions for dealing with char offsets in the drawing code - feel free to remove them if they are no longer used elsewhere.

Thanks for your work! 💪

aang521:
I think I fixed all the issues with multi byte characters.

src/widgets/text_input.jai
```diff
@@ -109,40 +108,25 @@ get_char_offset :: (bytes: [] u8, pos: s32) -> s32 {
 
 select_all :: (using input: *Text_Input) {
     cursor.pos = 0;
-    cursor.sel = get_num_chars(text);
+    cursor.sel = xx text.count;
 }
 
 select_word :: (using input: *Text_Input) {
-    // Decode the characters on the line for scanning
-    chars := decode_utf8_string_to_temp_chars(to_string(text), max_char = 5000);  // arbitrary number
```
obiwanus, line 115: So glad to see it gone - this function can finally retire.

obiwanus (changes requested):
Looks much better now! That temporary limit of 5000 has been bothering me and now we can get rid of it.

I think the only thing left to do it the `text_input_type_char` function, which still uses char offsets and causes crashes when you type non-ascii chars.

Would you be able to make that change? And delete the `decode_utf8_string_to_temp_chars` and `get_char_offset` while you're at it?

aang521:
Fixed the issue with typing non-ascii characters.
Also found and fixed some issues with deleting multibyte characters.

focus-editor:
Well done, thank you very much! 💪 

I noticed that the alt-delete shortcuts don't yet work, perhaps we should also add them later

PR #2, opened 2023-04-28
remember max_area to find biggest monitor

jlami:


focus-editor:
Good catch, thanks for the PR!

Issue #17, opened 2023-05-03
Assertion failed when switching projects


StevenSavold:
**Describe the bug**
When I open the editor in windows (I have not tried other OSes), via command line `.\focus_debug.exe` and then immediately hit `Ctrl+Shift+P` and select `Switch To Project` then picking any project I have in the given list, the program will hang for a few seconds, and then crash with the following assertion:

 F:/tools/Jai/jai-beta-1-062/modules/Thread/thread_group.jai:251,5: Assertion failed!
Stack trace:
F:/tools/Jai/jai-beta-1-062/modules/Preload.jai:380: default_assertion_failed
F:/tools/Jai/jai-beta-1-062/modules/Basic/module.jai:83: assert
F:/tools/Jai/jai-beta-1-062/modules/Thread/thread_group.jai:251: shutdown
L:/Dev/Third-Party/focus/src/widgets/finder.jai:16: deinit_finder
L:/Dev/Third-Party/focus/src/workspace.jai:145: hard_reload_workspace
L:/Dev/Third-Party/focus/src/main.jai:236: main
A breakpoint was hit, but no debugger is attached.

**To Reproduce**
Please see the description

**Expected behavior**
The program should successfully load the project and display the green notification in the bottom right

**Screenshots**
N/A

 - OS: Windows
 - Focus Version: Build from source with latest commit [commit hash: 1227d37c106051886bccebb1b8ffd8ac0e43a72c]

**Additional context**
It seems from my searching that the finder is being shutdown before it's initialized? Replacing [focus/src/widgets/finder.jai:16](https://github.com/focus-editor/focus/blob/main/src/widgets/finder.jai#L16) with `if finder.initted shutdown(*thread_group);` seems to fix the issue. But I'm unsure if the finder is expected to be initialized prior to this? If so the problem probably lies elsewhere. 

I have the fix in a fork if a PR is needed, but I feel this is such a small change its probably not worth the fuss? Let me know :)

focus-editor:
Hi, thanks, please send a PR and I'll accept it

PR #18, opened 2023-05-03
Fix for finder being shutdown before initialization when switching projects

StevenSavold:
This is the PR for issue #17


focus-editor:
Awesome, thanks for solving the issue

PR #19, opened 2023-05-04
README.md: add notes on MacOS compilation

meglio:


thisma:
I've seen similar CoreFoundation not found issues when Xcode _is_ installed. In that case a solution is to switch from Xcode to CommandLineTools. So, I'm not sure it makes sense to say Xcode must be installed. For now, I think this is good, just in case full Xcode is necessary. I wish I could try building it myself to figure it out.

meglio:
Unfortunately, I can't help here as I have not tested with other MacOS environments such as Intel-based MacBooks etc. I do agree that the comment I added might be overkill. Let's adjust it when we find the real cause of the problem.

Issue #20, opened 2023-05-04
Potential race condition when opening a new empty buffer? 


StevenSavold:
**Describe the bug**
It feels like there is a race condition when opening new empty buffers. When using Ctrl+N I will sometimes get a new empty buffer (expected), sometimes get a buffer with a random file from within my workspace directories (unexpected), and other times will get a crash (unexpected).

**To Reproduce**
1. Open the editor `.\focus.exe`
2. Open a new empty buffer `Ctrl+N`
3. If you get a valid empty buffer, type a few characters into the buffer and repeat from step 2 until you see the strange behavior

**Expected behavior**
A new empty buffer is created every time

**Screenshots**
N/A

 - OS: Windows
 - Focus Version: Latest commit (hash: 5e8ecee5c69198e741d4eca73d7d2e5d010c75d6)

**Additional context**
Here is the stack trace from when it crashes:
```
Array bounds check failed. (The attempted index is 0, but the array has dimension 0, so no index is valid). Site is L:/Dev/Third-Party/focus/src/buffer.jai:109.
Panic.
A breakpoint was hit, but no debugger is attached.

Printing the stack trace:
handle_exception                  F:\tools\Jai\jai-beta-1-062\modules\Runtime_Support_Crash_Handler.jai:363
... (skipping OS-internal procedures)
debug_break                       L:\Dev\Third-Party\focus\modules\Runtime_Support.jai:12
my_panic                          L:\Dev\Third-Party\focus\modules\Runtime_Support.jai:143
__array_bounds_check_fail         L:\Dev\Third-Party\focus\modules\Runtime_Support.jai:196
offset_to_coords                  L:\Dev\Third-Party\focus\src\buffer.jai:109
get_cursor_coords                 L:\Dev\Third-Party\focus\src\buffer.jai:125
draw_editor                       L:\Dev\Third-Party\focus\src\draw.jai:368
draw_frame                        L:\Dev\Third-Party\focus\src\draw.jai:21
main                              L:\Dev\Third-Party\focus\src\main.jai:228
```

The crash seems to be from the editor attempting to draw the buffer before the initialization code can add the first line to `line_starts`? But what is stranger to me, is when the buffer opens but is loaded with the content of a random file in my workspace directories.

StevenSavold:
I'd be happy to implement a fix, I just need more time to get to the bottom of what exactly is happening

focus-editor:
[...]

PR #23, opened 2023-05-04
Fix for "race condition when opening new empty buffer"

StevenSavold:
This is the PR for #20 

Let me know if you would rather have a different solution to the problem. Extending the critical sections like this is probably not ideal. 🤷‍♂️

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -478,15 +478,15 @@ get_line_as_string :: (using buffer: Buffer, line_num: s32) -> string {
 }
 
 find_or_create_an_empty_buffer :: () -> buffer_id: s64, created: bool {
+    lock(*open_buffers_lock); // We need the lock while searching because we might find an "empty" buffer while someone else is creating a new buffer
+    defer unlock(*open_buffers_lock);
+
     for * buffer, buffer_id : open_buffers {
         // Return the first empty standalone buffer to avoid creating a million new buffers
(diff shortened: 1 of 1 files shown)


src/buffer.jai
```diff
@@ -503,6 +503,7 @@ find_or_create_buffer :: (path: string) -> buffer_id: s64, created: bool {
         if path.count > 248 return -1, false;  // Windows bad
     }
     lock(*open_buffers_lock);
+    defer unlock(*open_buffers_lock);  // We need the lock potentially through setting the file data, because creating a new buffer will look for an "empty" buffer
```
focus-editor, line 506: This is a problem, it means that the workspace scanner can't add buffers in parallel any more. Yeah we need to find a different solution. I think maybe for now I will just allow to create a million new buffers if the user desires.

focus-editor:
I've added this commit which solves the problem - https://github.com/focus-editor/focus/commit/880836e2d3c51a345db3bb966379f2fc5d024405

Thank you for your help!

focus-editor:
Actually then I realised that the only thing that needed to be done was this https://github.com/focus-editor/focus/commit/51fa8bf5ce11987f1a5d4948da0b2988089b1d82

The flag is set only in the main thread so it's safe to check it

StevenSavold:
Happy to have been some help 😄

PR #24, opened 2023-05-04
Fix: macOS misc stuff

rluba:
These are most of the changes I made [on stream today](https://www.twitch.tv/videos/1810993894). Sorry for all the rants. It actually looks like a great editor. It’s just still very alien on macOS (for obvious reasons). 

The changes are quite a mixed bag of stuff.

I didn't put very much thought in the macOS key-bindings. They probably need to be changed a lot to match normal OS behavior, but the "keyboard combinations are not i18n-aware" issue prevented me from doing that. Not sure how we could tackle that issue.

Feel free to pick from these changes whatever fits your grand plan!


rluba:
No, wait. That probably doesn’t even compile because it relies on some changes I’ve made to compiler modules that aren’t yet in 0.1.062. I'll add those too.

ileonte:
This change in particular is going to come in handy for Linux as well: https://github.com/focus-editor/focus/pull/24/commits/0688fb8087389258a406770d03777739c116a081#diff-12aab456da5e695e671d6a845fb481ba54a2231001b1e8585736913ed5277b87

Right now there is a massive hack in the Linux backend where it just never generates `TEXT_INPUT` events when `Ctrl` or `Alt` are active, which is not correct. With this change we could handle these properly in the editor by choosing to ignore `TEXT_INPUT` events based on the value of `Event.text_input_count`.

focus-editor:
I will merge it and clean it up tomorrow - don't have the time right now so not merging just yet. Thanks Raphael!

PR #26, opened 2023-05-05
Support for per-platform config + custom scaling factor setting on Linux

ileonte:
This PR adds the following changes:
1. each platform backend now defines a top-level section for platform-specific options, a platform-specific structure to hold the options it cares about and a custom option parsing function that gets invoked from the main config parsing logic.
2. a new config file option on Linux to allow users to override the content scaling factor

### DIFF
--- a/src/config_parser.jai
+++ b/src/config_parser.jai
@@ -40,6 +40,9 @@ parse_config :: (name: string, filename: string, file_data: string) -> Config, s
                 success, error_msg := parse_style_line(*parser, line);
                 if !success return result, false, error_msg;
 
+            case .platform;
+                success, error_msg := platform_parse_config_line(*parser, line);
+                if !success return result, false, error_msg;
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/config_parser.jai
```diff
@@ -172,6 +177,11 @@ log_parser_error :: (handler: Text_File_Handler, format: string, args: .. Any, f
     return message;
 } @PrintLike
 
+log_parser_error :: (full_path: string, line_number: int, message: string, flags := Log_Flags.NONE, loc := #caller_location) {
+    new_message := tprint("Line % of '%': %", line_number, full_path, message);
+    log(new_message, flags=Log_Flags.ERROR|flags, loc=loc);
+}
+
```
focus-editor, line 184: Not sure I understand the purpose of this function - why is the one above not good enough?

ileonte: The purpose is to take a pre-formatted message as parameter instead of a format string. In theory such a message could contain a `%` character which would cause the existing version to fail.

focus-editor: Thanks for the explanation. I'll merge it now but please consider writing a short comment to not leave people wondering why it's there

PR #30, opened 2023-05-06
Fix input in the search bar not working on Linux (and probably MacOS as well)

ileonte:
Make `active_editor_handle_event()` aware of the latest input event handling changes.
Untested on Windows.


focus-editor:
My oversight, thank you for fixing it!

PR #41, opened 2023-05-14
Implement horizontal scrolling

ileonte:
Supports both keyboard and mousewheel based scrolling (only `shift+vertical-scroll-wheel` for now, will implement actual horizontal scroll wheel support in a future PR).

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -280,6 +280,11 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
 
     editor_is_active := editor_id == editors.active && Input.input_application_has_focus;
 
+    if editor.viewport.rect != rect {
+        editor.viewport.rect = rect;
+        editor.viewport.max_line_pixel_width = 0;
+    }
+
     // Calculate text origin in screen coordinates
     text_offset := make_vector2(char_x_advance, -char_x_advance);
     text_origin := top_left(rect) + text_offset - make_vector2(xx viewport.left, line_height - viewport.top);
@@ -416,6 +421,8 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
 
     // Draw editor
     {
+        min_visible_chars :: 5;
+
         push_scissor(rect);
         defer pop_scissor();
 
@@ -435,8 +442,8 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
             }
         } else if cursor_moved {
             // Detect if cursor is off screen and start scrolling to it
-            bounds := shrink_x(rect, text_offset.x + char_x_advance * 5);
-            bounds  = shrink_y(bounds, -text_offset.y + line_height * 5);
+            bounds := shrink_x(rect, text_offset.x + char_x_advance * min_visible_chars);
+            bounds  = shrink_y(bounds, -text_offset.y + line_height * min_visible_chars);
             bounds.h += 2;
             bounds.y -= line_height + 2;
             if bounds.w < 0 || bounds.h < 0 then bounds = rect;  // fall back
@@ -517,13 +524,17 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
             redraw_requested = true;
         }
 
-        // Smooth keyboard scrolling
-        if editor_smooth_scroll.active && editor_id == editors.active {
+        // Smooth scrolling - vertical
+        if editor_smooth_scroll.direction != .NONE && editor_id == editors.active {
             speed := 60 * line_height;  // pixels per second
             if editor_smooth_scroll.fast then speed *= 2;
             time_elapsed := cast(float) (frame_time - editor_smooth_scroll.started_at);
-            new_scroll_pos := editor_smooth_scroll.start_y + time_elapsed * speed * (ifx editor_smooth_scroll.up then -1 else 1);
-            viewport.top = clamp(cast(s32) new_scroll_pos, 0, max_y_scroll);
+            new_scroll_y := cast(float) editor_smooth_scroll.start_y;
+            if editor_smooth_scroll.direction == {
+                case .UP;    new_scroll_y -= time_elapsed * speed;
+                case .DOWN;  new_scroll_y += time_elapsed * speed;
+            }
+            viewport.top = clamp(cast(s32) new_scroll_y, 0, max_y_scroll);
             viewport.scroll_y.target = viewport.top;
             viewport_remember_glue_point(*viewport, buffer);
             redraw_requested = true;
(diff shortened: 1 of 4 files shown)


src/editors.jai
```diff
@@ -146,12 +148,12 @@ active_editor_handle_event :: (event: Input.Event, action: Action_Editors) -> ha
...
-        case .scroll_viewport_left;             // print("scroll_viewport_left - not implemented\n");
-        case .scroll_viewport_right;            // print("scroll_viewport_right - not implemented\n");
+        case .scroll_viewport_up;               smooth_scroll(editor, direction = .UP);                activate_hold_action(action, event);
+        case .scroll_viewport_up_fast;          smooth_scroll(editor, direction = .UP, fast = true);   activate_hold_action(action, event);
+        case .scroll_viewport_down;             smooth_scroll(editor, direction = .DOWN);              activate_hold_action(action, event);
+        case .scroll_viewport_down_fast;        smooth_scroll(editor, direction = .DOWN, fast = true); activate_hold_action(action, event);
+        case .scroll_viewport_left;             smooth_scroll(editor, direction = .LEFT);              activate_hold_action(action, event);
+        case .scroll_viewport_right;            smooth_scroll(editor, direction = .RIGHT);             activate_hold_action(action, event);
```
focus-editor, line 156: I have tried it and it works very well! However, even though the speed is the same as with a non-fast vertical scroll, it feels much slower, I suppose because the amount of information being scrolled into the viewport is much less - individual characters from some lines, as opposed to whole new lines. 

I would suggest that we use `fast = true` on these. I've tested it and with that option the scroll speed now feels the same both vertically and horizontally.

src/draw.jai
```diff
@@ -517,13 +524,17 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
             redraw_requested = true;
         }
 
-        // Smooth keyboard scrolling
-        if editor_smooth_scroll.active && editor_id == editors.active {
+        // Smooth scrolling - vertical
```
focus-editor, line 527: Please could you add a comment why we had to split those scrollings by direction, otherwise we will forget almost inevitably.

Also I added a comment below that we should indicate where we scroll by mouse and where by keyboard.

src/draw.jai
```diff
@@ -554,6 +565,44 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
...
+            line := array_view(buffer.bytes, line_start, length);
+            line_pixel_width := cast(s32) ((get_num_chars(line) + min_visible_chars) * char_x_advance);
+            viewport.max_line_pixel_width = max(viewport.max_line_pixel_width, line_pixel_width);
+        }
+        viewport_width := cast(s32) rect.w;
+        max_x_scroll := cast(s32) 0;
+        if viewport.max_line_pixel_width > viewport_width
+            max_x_scroll = viewport.max_line_pixel_width - viewport_width;
+
+        if (ui.hot_last_frame == ui_id || ui.hot_last_frame == scrollbar_id) && mouse_pointer_is_within(rect) && mouse.scroll_x_delta != 0 && !dont_scroll_this_frame {
```
focus-editor, line 580: Given how many ways of scrolling we now have it really would help if we had a comment before each. E.g. here it's scrolling by mouse, and below on line 591 it says "Smooth scrolling - horizontal" - we should probably clarify that it means keyboard scrolling. 
Same about the vertical scroll section.

src/draw.jai
```diff
@@ -554,6 +565,44 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
...
+            line_start := get_line_start_offset(buffer, line_num);
+            length := get_line_start_offset(buffer, line_num+1) - line_start;
+            line := array_view(buffer.bytes, line_start, length);
+            line_pixel_width := cast(s32) ((get_num_chars(line) + min_visible_chars) * char_x_advance);
+            viewport.max_line_pixel_width = max(viewport.max_line_pixel_width, line_pixel_width);
+        }
+        viewport_width := cast(s32) rect.w;
+        max_x_scroll := cast(s32) 0;
+        if viewport.max_line_pixel_width > viewport_width
+            max_x_scroll = viewport.max_line_pixel_width - viewport_width;
```
focus-editor, line 578: This codebase does not use this style of `if`. If it's a one-statement `if`, it's usually put on one line with a `then`, otherwise curly braces are always used.

Sorry for being nitpicky, just want to establish a sort-of consistent style

src/draw.jai
```diff
@@ -280,6 +280,11 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
 
     editor_is_active := editor_id == editors.active && Input.input_application_has_focus;
 
+    if editor.viewport.rect != rect {
+        editor.viewport.rect = rect;
+        editor.viewport.max_line_pixel_width = 0;
+    }
```
focus-editor, line 286: Two things.

1. I'm not sure why this is needed - I tried commenting it out and couldn't see the difference with a split editor and not. Could you please explain? Also I have a feeling that we don't need to compare rects there, maybe only comparing width would suffice?
2. My understanding is that unless we resize the editor we will keep increasing the max line pixel width every time we encounter a larger line. The problem with this is that it will allow us to scroll right way past the visible lines sometimes, and I'd want to avoid that - when you're scrolling right you should only reach the visible line limit (unless your `viewport.left` is already past that limit, in which case just don't scroll but maintain the position - that is, no snapping back to the limit)

focus-editor (changes requested):
Very good work, thanks for submitting it. Sorry for a million comments.

src/draw.jai
ileonte: `viewport.rect` is leftover from a test I ran, I will remove it.

As for `max_line_pixel_width`:
- I found it extremely annoying when the editor dynamically decreased the max scroll size. I'm used to be able to set the horizontal scroll position and then scroll vertically to see what else in the file is "visible" at that horizontal offset (it's pretty helpful when you try to keep line lengths close to a max size). I don't think having the view automatically adjust horizontally is a good experience.
- The fact that it doesn't currently have a way to decrease is something I think could be solved by adjusting it during file edit, but I didn't have the time to look into it: the idea is that we would need to remember the line number of the longest line; then, when processing input, if we modify the longest line in any way we just reset `max_line_pixel_width` to `0` and let the draw function recompute it.

focus-editor: > I don't think having the view automatically adjust horizontally is a good experience.

That's what I also said. If you scroll down after having scrolled to the end of a long line which is no longer visible, it definitely shouldn't snap back to the limit. But after you scroll to the left then the limit should shrink back, not allowing you to scroll right.

If you don't want to implement it, we can merge this and I can have a look later

focus-editor: I'm happy with the changes you've already made, so can merge and then we will look into shrinking it later

ileonte: It's not a question of not wanting to implement it, I'm just not sure I understand what the behavior you want is exactly. I just implemented the behavior I'm used to.

PR #44, opened 2023-05-19
Font handling tweaks for Windows

ileonte:
* `platform_get_fonts_dir()` actually attempts to get the fonts directory from the OS instead of simply returning a hard-coded value
* implemented `platform_find_font_by_name()`

### DIFF
--- a/src/platform/windows.jai
+++ b/src/platform/windows.jai
@@ -229,7 +229,6 @@ REFRESH_TIMER_ID :: 0;
 user32   :: #system_library "user32";
 kernel32 :: #system_library "kernel32";
 shell32  :: #system_library "shell32";
-gdi      :: #system_library "gdi32";
 ole32    :: #system_library "ole32";
 
 KF_FLAG_CREATE :: 0x00008000;                                       // https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/ne-shlobj_core-known_folder_flag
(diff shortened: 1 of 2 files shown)


src/platform/windows.jai
```diff
@@ -140,11 +140,60 @@ platform_get_save_file_name :: (name := "") -> string /* temp */, success: bool
...
-    return "C:/Windows/Fonts";
+    if !fonts_dir {
+        path_utf16: *u16;
+        ret := SHGetKnownFolderPath(*FOLDERID_Fonts, KF_FLAG_CREATE, null, *path_utf16);
+        defer CoTaskMemFree(path_utf16);
+        if SUCCEEDED(ret) {
+            fonts_dir = copy_string(wide_to_utf8(path_utf16));
+            for 0..fonts_dir.count - 1 {
+                if fonts_dir[it] == #char "\\" then fonts_dir[it] = #char "/";
+            }
```
focus-editor, line 191: For this we use `path_overwrite_separators(fonts_dir, #char "/")`

src/platform/windows.jai
```diff
@@ -140,11 +140,60 @@ platform_get_save_file_name :: (name := "") -> string /* temp */, success: bool
...
+                
+                ret := RegGetValueA(hkey, key.data, value.data, RRF_RT_REG_SZ, null, buff.data, *size);
+                if ret != 0 continue;
+                
+                path := to_string(buff.data, size);
+                if !is_absolute_path(path) {
+                    sb: String_Builder;
+                    init_string_builder(*sb);
+                    print_to_builder(*sb, "%/%", platform_get_fonts_dir(), path);
+                    return true, builder_to_string(*sb, allocator);
```
focus-editor, line 172: I'm not sure why a string builder is used here, since it only prints once.

How about something like this instead:
```
full_path := tprint("%/%", platform_get_fonts_dir(), path);
return true, copy_string(full_path, allocator);
```

src/platform/windows.jai
```diff
@@ -140,11 +140,60 @@ platform_get_save_file_name :: (name := "") -> string /* temp */, success: bool
...
+    // not monospace. However it only enumerates fonts that have been loaded in GDI via
+    // `AddFontResourceEx()` which doesn't include fonts installed in the user specific
+    // location (`%LocalAppData\Microsoft\Windows\Fonts`) by default.
+    //
+    // Instead of spamming `AddFontResourceEx()` calls we just take file names straight
+    // from he registry.
+
+    hkeys  : []HKEY   = .[HKEY_CURRENT_USER, HKEY_LOCAL_MACHINE];
+    styles : []string = .[" Regular", ""];
+    types  : []string = .[" (TrueType)", ""];
```
focus-editor, line 153: The FreeType library which we use for rendering supports OpenType fonts as well - should we add it here too?

src/platform/windows.jai
```diff
@@ -182,12 +233,73 @@ REFRESH_TIMER_ID :: 0;
...
+    tmStruckOut: u8;
+    tmPitchAndFamily: u8;
+    tmCharSet: u8;
+}
+
+DEFAULT_CHARSET :: 1;
+FIXED_PITCH :: 1;
+
+FONTENUMPROCA :: #type (elfe: *ENUMLOGFONTEXA, me: *TEXTMETRICA, FontType: DWORD, lParam: LPARAM) -> s32 #c_call;
+EnumFontFamiliesExA :: (hdc: HDC, lf: *LOGFONTA, proc: FONTENUMPROCA, lParam: LPARAM, dwFlags: DWORD) -> s32 #foreign gdi;
```
focus-editor, line 302: If we're not using `EnumFontFamiliesExA`, perhaps we should remove all the structs and stuff? 

Or are you planning to use it later?

focus-editor (commented):
Awesome!

src/platform/windows.jai
ileonte: Seems like Windows "tags" OpenType fonts as `(TrueType)` as well:
image
This is what I tested with, seemed to work fine.

src/platform/windows.jai
ileonte: The first version I wrote used `EnumFontFamiliesExA()` (there's a comment about it in `platform_find_font_by_name()`). Since it's not part of the official `Windows` module I thought I'd just leave the bindings in place in case we do need them in the future. But I can remove them if you want.

focus-editor: I'd suggest to remove them for now because they are big and distracting - we do have them in this PR to copy from if we want to use them in the future, so your work isn't lost.

PR #46, opened 2023-05-19
Keybindings with multiple key combos

arjenwitteveen:
This is an implementation for keybindings consisting of more than one key combo (e.g., binding `Ctrl-X Ctrl-S` to a specific command). I have resorted to the following nomenclature: a keybinding is defined by a *key sequence*, which consists of one or more *key combos*. Currently, all commands in Focus are called using key sequences consisting of just one key combo. Something like `Ctrl-X Ctrl-S` would be an example of a key sequence consisting of two key combos. This type of key sequence is very prevalent in Emacs, but also exists in editors like VS Code (I think they call it 'key chords' there).

I've tested the implementation on Windows and macOS.

#### `src/keymap.jai`, `src/main.jai`
These changes implement the actual key sequence handling. A global variable of type `Key_Sequence_Input_State` keeps track of whether a key sequence is in progress, and if so, which `Key_Mappings` are a match for the key sequence so far. Some state is recorded per frame to know if any relevant input occurred, if a complete match was found this frame, and what the new matches are. An initial key combo will be matched against the entire keymap; further combos will be matched only against the list of current matches. The key sequence ends when a complete match has been found, or when no matches remain.

---

#### `src/config_parser.jai`
The parser has been updated to support key sequences. The format of the config is the same as before, except that you can now have N key combos in front of an action rather than just one.

One subtlety that I ran into: since config files are merged (e.g., the global and the default configs), there is a risk that a config higher up in the hierarchy will define a keybinding that is the same as the prefix of a custom keybinding (e.g., we have a custom `Ctrl-X Ctrl-S` keybind, but the default config will define `Ctrl-X` for cut). I worked around this by saying that when (part of) a keybinding is matched, key sequences shorter than the one that was already matched will no longer be considered. In our example, when pressing `Ctrl-X`, our `Ctrl-X Ctrl-S` keybind will be encountered first; since it consists of 2 key combos, from that point onwards only key sequences with at least 2 key combos will be considered, so the `Ctrl-X` from the default config will be ignored. This seems like a decent way to deal with collisions between different configs. Perhaps a future improvement could be to warn users when there are keybindings that match the prefix of another keybinding, at least within a single config file.

---

#### `src/editors.jai`, `src/widgets/text_input.jai`
Added checks to prevent text input while a key sequence is in progress.

---

#### `src/draw.jai`
Edits to draw shortcut keys correctly in various places. I also added a bonus feature: while a key sequence is in progress, the key sequence so far will be displayed in the status bar (for key sequences that consist of one part, nothing will be displayed). It's just a suggestion though, maybe some will find it annoying.

---

#### `modules/Input/macos.jai`
Not totally necessary for this feature, but this change fixes an issue where the righthand side modifiers keys were not handled properly on macOS, causing key sequences not to always work with those keys. I sent in a bug report for this, so it will probably be fixed in one of the coming betas.

---

Some other points to consider:

- `parse_keymap_line` in `src/config_parser.jai` creates a resizable array to store each key sequence; those arrays use the default allocator right now (they ultimately end up in a `Key_Mapping` in a keymap). I don't think this is a big deal since parsing configs won't happen often in the general case, but perhaps the memory management could be cleaner. I considered a fixed size array on `Key_Mapping` and have key sequences have a max length (say 4 or 8), but that seems uglier.

- It is possible to use optional modifier syntax anywhere in a key sequence (e.g. `Ctrl-X {Ctrl}-S`); this seems to work as expected, but there may be some edge cases where it doesn't as I haven't tested it thoroughly.

- I decided that hold actions should not apply to key sequences with more than one key combo, since that doesn't seem to make that much sense.

### DIFF
--- a/modules/Input/macos.jai
+++ b/modules/Input/macos.jai
@@ -9,112 +9,116 @@ window_minimized := false; // @@ Hack.
 osx_externally_generated_events: [..] Event;
 
 set_custom_cursor_handling :: (is_custom: bool) {
-     // @Incomplete: Implement this.
-     // Use this if you are doing immediate-mode API stuff that sets cursors and want to prevent cursors from flickering. Windows is annoying to interface with.
+    // @Incomplete: Implement this.
+    // Use this if you are doing immediate-mode API stuff that sets cursors and want to prevent cursors from flickering. Windows is annoying to interface with.
 }
 
 #scope_file
 
-OSX_VK_ESCAPE     :: 53;
-OSX_VK_F1         :: 122;
-OSX_VK_F2         :: 120;
-OSX_VK_F3         :: 99;
-OSX_VK_F4         :: 118;
-OSX_VK_F5         :: 96;
-OSX_VK_F6         :: 97;
-OSX_VK_F7         :: 98;
-OSX_VK_F8         :: 100;
-OSX_VK_F9         :: 101;
-OSX_VK_F10        :: 109;
-OSX_VK_F11        :: 103;
-OSX_VK_F12        :: 111;
-OSX_VK_F13        :: 105;
-OSX_VK_F14        :: 107;
-OSX_VK_F15        :: 113;
-OSX_VK_BACK_TICK  :: 50;
-OSX_VK_1          :: 18;
-OSX_VK_2          :: 19;
-OSX_VK_3          :: 20;
-OSX_VK_4          :: 21;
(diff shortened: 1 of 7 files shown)


focus-editor (approved):
Thank you for your work. I will merge these changes now and will test them some more before the next release. We will need to change the way the default config is handled before these go live.

PR #47, opened 2023-05-19
Add case sensitive and whole word toggleable search options

hfr4:
Toggle search options via keyboard shortcut when search boxes are open (Alt + C / Alt + W by default like VSCode)
Toggle via mouse click on the "buttons" is not implemented.


focus-editor (commented):
I've tested the changes and they work as expected, thanks.

I would still like to make a few adjustments. I could merge your changes and then make the adjustments myself if you don't mind.

focus-editor:
Thanks - I will make the changes before the next release!

PR #49, opened 2023-05-22
open_file_in_explorer command

cookednick:
- Added the command to keymap.jai and open_file_dialog_handle_event
- Added it to editors section of default configs as Shift-Enter
- Modified platform_open_in_explorer to accept an argument on whether to reveal a file/path or open it (open = the default)
- Modified draw.jai to support Shift+Click both on path chunks and entries in the navigate dialog
- Added a few special-cases to trim (or add) trailing slashes in order to make this work ("C://thing" -> "C:/thing" & "C:" -> "C:/" on Windows, "//" -> "/" on Mac/Linux)

Extensively tested. Everything I tried worked on both Windows and macOS, including files/directories with spaces in their paths. Seems to be solid. Even fixed a few bugs that I discovered. (These bugs did not seem to cause problems elsewhere in the program. Oh well. They are now fixed.)

### DIFF
--- a/default_macos.focus-config
+++ b/default_macos.focus-config
@@ -140,6 +140,7 @@ Enter                       open_entry_in_place
 Cmd-Enter                   open_entry_on_the_side
 Cmd-1                       open_entry_on_the_left
 Cmd-2                       open_entry_on_the_right
+Shift-Enter                 open_entry_in_explorer
 
 Tab                         open_directory
 
@@ -155,7 +156,7 @@ Cmd-Enter                   open_entry_on_the_side
 Cmd-1                       open_entry_on_the_left
 Cmd-2                       open_entry_on_the_right
 
-Shift-Enter                 move_up  # an Cmdernative way to move
+Shift-Enter                 move_up  # an alternative way to move
 
 
 [common]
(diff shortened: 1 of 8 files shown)


default_macos.focus-config
```diff
@@ -155,7 +156,7 @@ Cmd-Enter                   open_entry_on_the_side
 Cmd-1                       open_entry_on_the_left
 Cmd-2                       open_entry_on_the_right
 
-Shift-Enter                 move_up  # an Cmdernative way to move
+Shift-Enter                 move_up  # an alternative way to move
```
focus-editor, line 159: Loooool

src/widgets/open_file_dialog.jai
```diff
@@ -130,6 +131,30 @@ open_file_dialog_open_entry :: (selected: s64, placement: Editor_Placement, fold
...
+    entry := entries.filtered[selected];
+    
+    path: string = ---;
+    #if OS == .WINDOWS {
+        if ends_with(entry.name, ":") {
+            // Drive letter. We need to add a slash.
+            b: String_Builder;
+            append(*b, entry.name);
+            append(*b, #char "/");
+            path = builder_to_string(*b, allocator = temp);
```
focus-editor, line 146: Why not `tprint`? Would be one line.

focus-editor (approved):

src/widgets/open_file_dialog.jai
cookednick: You're right. I was very tired. Fixing.

PR #50, opened 2023-05-27
partial fix for large text pasted from other windows on Linux

ivan-cx:
Problem:
Current version has a limit of 128 bytes for texts pasted from other windows on Linux.

Reason:
in modules/Linux_Display/ldx_input.jai line 365
there is length 32 in the arguments AND later remainder bytes (*rem) return value is ignored

Partial fix:
Send ~0 instead of 32 to remove the upper limit on the length of one chunk. So, all bytes are sent in one huge chunk.
However, it does not handle really huge texts yet.
Also added assert(rem == 0) and check for return data type and format.

P.S. thanks for the cool editor! Really enjoying it so far

### DIFF
--- a/modules/Linux_Display/ldx_input.jai
+++ b/modules/Linux_Display/ldx_input.jai
@@ -362,13 +362,18 @@ x11_clipboard_get_text :: (win: Window) -> string {  // Allocated via alloc; sho
         format: s32;
         data: *u8;
         type: Atom;
-        if XGetWindowProperty(display, win, property, 0, 32, True, AnyPropertyType, *type, *format, *nitems, *rem, *data) == 0 {
-            temp: string;
-            temp.data = data;
-            temp.count = xx nitems;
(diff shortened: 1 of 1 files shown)


mathaou:
This is very helpful! I encountered this issue multiple times but just never got around to filing an issue for it.

ivan-cx:
For future reference: here is the code for Linux console program **xsel** that outputs contents of the clipboard to the console and handles that huge **INCR** transfers: https://github.com/kfish/xsel/blob/0ee15bfc8139d8eb595004ea093c6a57b7134470/xsel.c#LL731C27-L731C29

ileonte:
Thanks for fixing this. You should also report the problem on the Jai beta discord, this code is copy-pasted from the `Clipboard` module.

ivan-cx:
@ileonte thank you! 
yes, I've already reported this privately via e-mail

PR #53, opened 2023-06-01
Add C# syntax highlighting

audV:
Started with c.jai as a base.

Parsing @, $, number, character and string literals are probably where the most difference is to c.jai as C# has a lot of variation for those.

Invalid tokens are used explicitly in a couple of cases where tokens are obviously incorrect. Other highlighters didn't do this much or at all so I didn't go too far with it.

Cleaned up the compile time hash tables at the bottom a little by removing the need for #insert -> string

Changed the assert at the bottom to not use enum_highest_value as this could be incorrect.

There will be some slight coloring differences to Visual Studio highlighting. For example there is no color for access modifiers in Focus. In some cases I opted for coloring that was more Focus oriented like coloring types differently to keywords whereas VS does not do this.

Issues discovered in other highlighters
- last_token appears to be used before it is initialized.
- parse_number does not handle when a '.' is the second character (colored as an operator I think).

Not Done
- Interpolated string parsing (this involves nested parsing (tokens in tokens) as it seemed like it might require some architectural consideration/change). They are just handled as normal strings without any interpolated identifier highlighting. Verbatim strings are handled, as well as combinations of verbatim/interpolated.
- Summaries /// are just handled the same as comments without any additional highlighting.

I have C# test solution that I can send over if it will help with testing

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -18,7 +18,7 @@ __pycache__
 # These files are explicitly allowed so that the editor doesn't waste time figuring out
 # whether it should ignore these or not (if a file is not explicitly allowed or ignored,
 # it will be read to determine if it's binary or not).
-.jai .c .cpp .h .hpp .cc .txt .md .focus-* .ini .csv .log .sql .py .m .html .xml .plist .js .ts .yml .yaml .toml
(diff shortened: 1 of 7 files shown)


focus-editor:
Thanks!

PR #54, opened 2023-06-02
Fixed select_word, Added center_viewport_on_cursor, Added editor history

onelivesleft:
* Fixed `select_word` editor binding not working.
* Added `center_viewport_on_cursor` command.
* Added `dump_key_mappings` command (dumps all key mapping to console for debugging).
* Added rudimentary editor history feature, let's you move back and forth through your recent edit locations.

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -85,10 +85,10 @@ active_editor_handle_event :: (event: Input.Event, action: Action_Editors) -> ha
             case .toggle_expand;                    search_bar_toggle_expand(editor);                                             return true;
             case .toggle_case_sensitive;            search_bar_toggle_case_sensitive(editor, buffer);                             return true;
             case .toggle_whole_word;                search_bar_toggle_whole_word(editor, buffer);                                 return true;
-            
+
             case .search_in_buffer;                 open_search_bar(editor, buffer, remember_cursor_position = false);            return true;
             case .search_in_buffer_dropdown_mode;   open_search_bar(editor, buffer, .dropdown, remember_cursor_position = false); return true;
-            
+
             case .move_up;                          search_bar_move_cursor(editor, buffer, -1, wrap = true);                      return true;
             case .move_down;                        search_bar_move_cursor(editor, buffer,  1, wrap = true);                      return true;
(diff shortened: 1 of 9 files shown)


focus-editor (commented):
Thanks for submitting the PR @onelivesleft !

Currently this will crash the editor in certain circumstances, e.g. when you try to rewind to a position which no longer exists. To reproduce:
- Move the cursor around
- Delete all file contents
- Try to rewind

I understand this is not meant to be finished-finished, so I could either take care of that myself when I finally have the time, or you could adjust it if you want.

onelivesleft:
I'll fix it.  I'm almost done the fancier system where it compresses moves down by action; wasn't sure if you'd want to build your own off this simple one (like the timer you mentioned) ; do you want that?

focus-editor:
No the timer was just an idea - the system you proposed sounds interesting, it might work better. Thanks!

src/widgets/commands.jai
```diff
@@ -186,6 +189,8 @@ commands := #run Command.[
 
     .{ .move_editor_to_the_left,                            "Move Pane To The Left",              0, .Double },
     .{ .move_editor_to_the_right,                           "Move Pane To The Right",             0, .Double },
+
+    .{ .dump_key_mappings,                                  "Dump Key Mappings",                  0, .None },
```
focus-editor, line 193: I wouldn't expose it in release builds because it could be confusing to some people, can we `#if DEBUG` it maybe?

src/editors.jai
```diff
@@ -2306,6 +2351,7 @@ Editor :: struct {
     last_cursor_created_using_mouse_at: Time;
 
     scroll_to_cursor: enum u8 { no; yes; yes_new_editor; }  // trigger a scroll so that the cursor is in the center vertically
+    pane_location: enum u8 { single; left; right; }
```
focus-editor, line 2354: Seems to be an unused leftover

onelivesleft:
Fixed the above

focus-editor (commented):
Thanks - sorry, been busy all day and couldn't finish testing/reviewing - will finish tomorrow.

PR #56, opened 2023-06-05
Copy position and cursors when splitting editor

hfr4:
Previously when duplicating an editor the cursor was placed at the top of the file, now it open at the same position and all the cursors positions are copied.

https://github.com/focus-editor/focus/assets/30574115/0a7dd93b-c8f3-4940-b6de-c53faa41ab9d


focus-editor:
Awesome, thanks

PR #57, opened 2023-06-06
Added scroll_to_cursor_line_offset setting

onelivesleft:
Adds a line offset when centering view on cursor.  Default value is 1, which compensates for the slight too-low bias of the centering math caused by status bar existence.

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -53,6 +53,7 @@ tab_size:                               4
 line_height_scale_percent:              120
 max_editor_width:                       -1
 editor_history_size:                    128
+scroll_to_cursor_line_offset            1
 
 
 [[keymap]]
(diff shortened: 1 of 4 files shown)


default.focus-config
```diff
@@ -53,6 +53,7 @@ tab_size:                               4
 line_height_scale_percent:              120
 max_editor_width:                       -1
 editor_history_size:                    128
+scroll_to_cursor_line_offset            1
```
focus-editor, line 56: I'm a bit reluctant to add settings which most people will not want to use. Why don't we just hard-code this `1`?

focus-editor:
Not sure why you closed it - I wouldn't mind to add this offset, just not sure why it would ever need to be anything other than 1. Happy to accept a PR with a hard-coded 1

PR #60, opened 2023-06-07
add copy_line_without_selection setting

hfr4:
New setting to copy full line if there is no selection, currently work when copying/pasting with 1 cursor only.

I also separated editor cut/copy_to_clipboard functions to simplify control flow

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -48,12 +48,12 @@ strip_trailing_whitespace_on_save:      false
 smooth_scrolling:                       true
 double_shift_to_search_in_workspace:    false
 can_cancel_go_to_line:                  true
+copy_line_without_selection:            false
 max_entries_in_open_file_dialog:        2000
 tab_size:                               4
 line_height_scale_percent:              120
 max_editor_width:                       -1
 editor_history_size:                    128
-
 
 [[keymap]]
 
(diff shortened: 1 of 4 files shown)


focus-editor (commented):
This seems overcomplicated for what it's doing. I had a feeling that all that was required was just a flag and a single line change. Are you sure it can't be achieved by that?

I'm going to try implementing it myself and see first.

focus-editor:
I've added this commit https://github.com/focus-editor/focus/commit/038fde876812c380a9801602144542a0637e9bd8

It will have a side effect of selecting the line when copying, which presumably you were trying to avoid, but it has a bunch of nice features when done this way:
- it will work with multiple cursors as expected (including when the cursors are on the same line)
- it has a clear indication that something happened when you press `copy` - I was afraid it would be annoying, but it wasn't when I tried it
- the code is much more straightforward and doesn't require any extra state per editor

focus-editor:
Given that we're copying the line, we're probably going to paste it somewhere else, so the cursor position within the line is not as important as in other cases.

That said, if a comparably simple solution is found which doesn't select the line, but works for multiple cursors and doesn't need an extra piece of state per editor, I'll be happy to reconsider.

> It still have weird behaviors with multi cursor when pasting (insert new lines when pasting / if you reset to one cursor then to the previous number of cursors the clipboard is merged)
This needs to be considered separately.

PR #61, opened 2023-06-07
Fixed the multi-line copying and pasting

polgartom:
https://discord.com/channels/661732390355337246/1100701456950099978/1103153377237540874

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -2199,7 +2199,6 @@ copy_selection_to_clipboard :: (editor: *Editor, buffer: *Buffer, cut := false) 
     for * cursor : editor.cursors {
         str := get_selected_string(cursor, buffer);
         append(*b, str);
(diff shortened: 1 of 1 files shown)


polgartom:
I realized this is was stupid because I don't know what is the desired behavior. What we want like vscode?

focus-editor:
I believe removing that line would cause issues elsewhere. We need to have a closer look at it.

We do want the behaviour to be similar to that of vscode, but not 100% because vscode seems to treat cursors created by ctrl-clicking and cursors created by ctrl-D differently, and we don't want that complexity

polgartom:
Got it! Thanks. Yes it will cause a weird effect. This is why I asked what we do want. The issue will appear if you copy with multiple cursor without selected range, in that case I believe every line will get an extra "\n".

PR #62, opened 2023-06-08
add function call and type name highlight for jai

hfr4:
Add more syntax highlight for jai files


src/config.jai
```diff
@@ -424,7 +428,9 @@ refresh_code_color_map :: () -> [Code_Color.NUM_COLORS] Vector4 {
     result[DEFAULT]         = Colors.CODE_DEFAULT;
     result[COMMENT]         = Colors.CODE_COMMENT;
     result[TYPE]            = Colors.CODE_TYPE;
+    result[TYPE_NAME]       = Colors.CODE_TYPE_NAME;
     result[FUNCTION]        = Colors.CODE_FUNCTION;
+    result[FUNCTION_CALL]   = Colors.CODE_FUNCTION_CALL;
```
focus-editor, line 433: I sort of understand that some people might want to color function definitions and function calls differently, and color struct/enum/union keywords differently as well, but I'm hesitant to merge this because the naming is a bit confusing.

E.g. people will be wondering what is `code_type` and `code_type_name` and what the difference is. Also `code_function` should then be named `code_function_definition` or something, but renaming config options right now is something to be avoided, at least until we have a config rewrite system in place, which @cookednick said he wants to write.

cookednick: I'll prioritize that and get it done in the next two weeks then. Then we can merge this (with maybe those changes) as part of that new upgrade system.

PR #64, opened 2023-06-10
Fixed multi-line copying and pasting bug & Fixed C #endif syntax highlight & Added #ifdef directive

polgartom:
This issue listed in the TODO.md file.
https://discord.com/channels/661732390355337246/1100701456950099978/1103153377237540874

### DIFF
--- a/run.bat
+++ b/run.bat
@@ -1,3 +1,3 @@
 @echo off
 
-jai -quiet -import_dir modules first.jai && .\focus_debug.exe
+jai -quiet -import_dir modules first.jai && .\focus_debug.exe src/config.jai
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -2199,7 +2199,9 @@ copy_selection_to_clipboard :: (editor: *Editor, buffer: *Buffer, cut := false) 
     for * cursor : editor.cursors {
         str := get_selected_string(cursor, buffer);
         append(*b, str);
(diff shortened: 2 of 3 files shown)


polgartom:
I hate this github stuff. Just copy my changes manually if you want to use it, just only few line of code

focus-editor:
https://github.com/focus-editor/focus/commit/bb19fc1480f8397bb0ad8e424e552946f98c0b7b - slightly changed to make sure the last cursor doesn't get a trailing newline

https://github.com/focus-editor/focus/commit/b2f0b359f0184fa08192bd361e3cd8c2153829bc#diff-9c3cf967631ecc2f7398fd3e765e7f857b6fcc2453de84e28dcb49a39e17f69bR668

I didn't understand the other change you made, so I didn't include it

polgartom:
I didn't understand either. The change what I made was I added two missing directives to C syntax highlighting #ifdef and #ifndef. Also I changed a little bit the directive highlighting mechanism, because if let's say you have a c header file probably the last token in that file is the #endif directive. Now, it won't highlight, because you have to add whitespace next to the directive. (as you will see in the images) I'll create an another commit then.

image
image

focus-editor:
It's a general bug of the highlighter - any last token won't be recognised unless you add a newline or a space. I know about it, just never got around to fixing it yet

PR #68, opened 2023-06-15
Better clipboard handling

hfr4:
Correct a bug when copy pasting with the same number of cursors if you set them up yourself (inside the same editor or in two different editors)

Previous:

Now:

To allow this, clipboard is now global to the application and not contained in editor state same for "global_clipboard_slices" which represent what previously was  "cursor clipboard".

Overall reduction of code complexity for clipboard handling.

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -908,13 +908,11 @@ Buffer :: struct {
 
 Cursor :: struct {
     using state: State;
-    clipboard: Clipboard;
 
-    State     :: struct { pos, sel: s32; col_wanted: s32 = -1; }
-    Clipboard :: struct { start: s32; count: s32; }
+    State :: struct { pos, sel: s32; col_wanted: s32 = -1; }
 }
 
-#assert size_of(Cursor) == (2 + 1 + 2) * size_of(s32);  // make sure no padding accidentally gets added
+#assert size_of(Cursor) == (2 + 1) * size_of(s32);  // make sure no padding accidentally gets added
 
 Cursor_Coords :: struct {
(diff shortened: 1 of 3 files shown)


focus-editor (approved):
I like this change a lot.

May I ask you to fix a related bug before this pull request is merged?
To reproduce, create a couple of cursors without a selection and then another one with a selection. Then copy and paste - the last cursor's clipboard will be cut - see the attached gif.
focus

It's an existing bug so if you don't want to fix it I can merge as is.

focus-editor:
I'm sorry, I just tried it and it seems very broken now. Like, I tried copying a single word and pasting it back and it cut of the last symbol. 

I could merge the original PR if you want and then we could look into those other things, because the original PR worked very well

focus-editor:
Sorry for only getting to it now - was sick. 

I'm going to merge it now since it closely matches VSCode behaviour and that means it's the least surprising to people. I will however make a couple of adjustments after that:

1. There seems to be a variable `was_selected` which doesn't really seem to be used
2. I want to try to avoid rescanning for lines for each cursor, in case there's a way to do that.

Thanks for your work!

focus-editor:
I've managed to spend some time properly reviewing the changes and they actually seem OK, so I won't be making any changes unless issues are found 💪

focus-editor:
FYI cutting was completely broken (it would always delete whole lines). 

Don't know how I missed it when testing 🤦 

Fixed here https://github.com/focus-editor/focus/commit/5d28b4136e2bdf159d28b092d5ba5870155ff9b1

PR #70, opened 2023-06-17
Feature - Move between next and previous buffers

polgartom:
I missed this feature so much. Probably later we wanna jump between the buffers relatively to the last used ones.

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -202,6 +202,9 @@ Ctrl-Shift-W                close_other_editor
 
 Ctrl-Shift-L                toggle_expand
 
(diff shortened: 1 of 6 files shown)


focus-editor:
What's the reason for closing? I haven't looked at it yet because I'm busy.

polgartom:
This already exists in a similar way. So I don't want to bloat this good stuff with my stuff. I reopen it and try it if you want. Probably it has some sense to put it but I don't know at this point. Today I'll make an another PR then, also I found an "Array bounds check fail" problem in the Editors history feature. I'll take care of it.

PR #75, opened 2023-06-21
Open window on primary monitor

Mulej:


focus-editor:
Does it not open on primary by default if you have `open_on_the_biggest_monitor` set to `false`? If not, then we should probably fix that rather than adding another option which is in conflict with the other option.

The plan is to remember monitor position and open it where it was last time eventually.

Mulej:
> Does it not open on primary by default if you have open_on_the_biggest_monitor set to false?
No, for some reason windows enumerates my secondary monitor first. I will modify the pull to open on the primary monitor by default if you don't have `open_on_the_biggest_monitor` set.

> The plan is to remember monitor position and open it where it was last time eventually.
Yeah, I'm working on saving/restoring sessions right now. The plan is to save the position/size in the session and restore it when loading the session.

PR #76, opened 2023-06-22
Make it so that we will always parse the global config file on first load, even if it has errors.

Mulej:


focus-editor:
As I understand it the problem we're trying to solve is that no keymaps are loaded if the config file has issues. This has been introduced with the recent change where we don't mix stuff into the default config every time.

I wouldn't, however, fix it by trying to parse every line as a keymap line. While I understand that I would load some of the user key mappings, it would not work reliably in every case and could still leave the editor in an unusable state.

I think a more robust way to do it would be to use the default config instead if the first load doesn't succeed.

focus-editor:
I've added a fall back to default https://github.com/focus-editor/focus/commit/0d15175fda0842358482da4c752ca4d16c87b923. It is now possible to fix config errors. 

I understand that it will not load the user's preferred key mappings, but both solutions are not perfect, and this one is much more predictable.

PR #78, opened 2023-06-23
Buffer history - Move to previous and next buffer

polgartom:


focus-editor:
I don't have the time for a proper review right now, but what I'm worried about with a separate buffer history is how it will play out together with editor history that's already implemented. 

If I were implementing this I would reuse the existing editor history and would just skip all records within the same buffer when going back to previous/next buffer, preserving all the existing records naturally. As I can see here it has a separate array. Not sure if that's the best decision. 

I'll get back to it as soon as I can.

focus-editor:
Had another look at this. 
I would suggest that we try to use one history array for both cursor positions and editors as I mentioned instead of having another separate one.
There are some issues with the current editor history which need to be fixed first though - I'll get to it hopefully soon and then I could potentially cherry-pick some things from your PR to use in buffer history.

polgartom:
Yes, I guess you're right. Right now, I don't know what you can cherry-pick from this small piece of code. There is "Review the editor history implementation" in the TODO. If you don't mind, I will look into it. I also found an array bounds check error in editors.jai. I will address that as well.
This editor is deserve almost every free time of mine.

focus-editor:
I've actually already started looking into the editor history stuff - there's a few bugs that need to be fixed and also I wanted to change a couple of things that I found confusing.
That TODO is a leftover for my personal todo list. I should probably keep it away from the main codebase now.

polgartom:
Ohh, alright then!

focus-editor:
I finally was able to fix the issues and implement the buffer history - @polgartom  please have a look and check if it works as you would expect.

polgartom:
@focus-editor It's not working as I expected. The idea behind it was to have separate buffer histories for the left and right editors. What I did was to simple navigation between buffers (similar) to having tabs. I guess right now it remains the same as it was but with larger steps. I'm not sure how useful this is right now, you can probably remove it. Additionally I don't know how useful my version is for other people but it is very useful for me but you'll decide.

obiwanus:
It makes sense. We should consider adding it on top.

polgartom:
Another option is to add tabs to it which can be optional by selecting the "New editor tab" from commands or something just like in vim (or nvim idk)

PR #79, opened 2023-06-24
Add toggle_block_comment command

Mulej:


src/editors.jai
```diff
@@ -2354,6 +2356,101 @@ toggle_comment :: (editor: *Editor, buffer: *Buffer) {
...
+        for * cursor : editor.cursors {
+            move_count := cast(s32)it_index * comment_total_length;
+
+            start_pos := min(cursor.pos, cursor.sel) + move_count;
+            end_pos := max(cursor.pos, cursor.sel) + move_count + 2;
+
+            insert_string_at_offset(buffer, start_pos, comment_start);
+            insert_string_at_offset(buffer, end_pos,   comment_end);
+            cursor.pos += 2 + move_count;
+            cursor.sel += 2 + move_count;
```
focus-editor, line 2447: I'm sorry, I don't have the time right now to have a proper look - will do as soon as I can, but what's the purpose of this "2"?

Mulej: Ah, forgot to un-hard code the length of the comment start. Fixed now :)

src/editors.jai
```diff
@@ -2354,6 +2356,101 @@ toggle_comment :: (editor: *Editor, buffer: *Buffer) {
...
 
+toggle_block_comment :: (editor: *Editor, buffer: *Buffer) {
+    comment_start: string;
+    comment_end:   string;
+
+    if buffer.lang == {
+        case .C;            comment_start = "/*"; comment_end = "*/";
+        case .CSharp;       comment_start = "/*"; comment_end = "*/";
+        case .Jai;          comment_start = "/*"; comment_end = "*/";
+        case .Focus_Config; toggle_comment(editor, buffer); // The config doesn't have block comments, fall back to toggle_comment.
```
focus-editor, line 2367: I wonder if if would make sense to always fall back to `toggle_comment` for all languages?

src/editors.jai
```diff
@@ -2354,6 +2356,101 @@ toggle_comment :: (editor: *Editor, buffer: *Buffer) {
...
+
+        end_search_index += selection.end;
+
+        array_add(*indices, start_search_index);
+        array_add(*indices, end\ _search_index);
+
+        for color_index : start_search_index..end_search_index {
+            remove_comment &= buffer.colors[color_index] == xx Code_Color.COMMENT;
+            if !remove_comment then break cursor;
+        }
```
focus-editor, line 2403: I don't think we should depend on buffer coloring. Those systems are independent and it creates an unnecessary hard link between the two.

We are planning to change the way colors are assigned in the near future, there will not be a color per byte any more.

focus-editor (changes requested):
Thanks for your PR. I'd like to merge it but there are some issues.

1. I left a comment about depending on buffer colors - we really shouldn't do that.

2. Comment detection seems to be broken when you try to comment within a string (it doesn't work at all if you only do it with one cursor and not multiple)
focus

src/editors.jai
Mulej: Yeah, I'm using it to detect if we're in a comment to switch between adding them and removing them. How would you feel about saving an array of tokens from the tokenizer in the buffer for this kind of thing? It would also be useful for other things, like goto definition.

src/editors.jai
Mulej: That makes more sense, will do.

src/editors.jai
focus-editor: What problems would there be if we just searched from the edges outwards until we either found the enclosing comments or we didn't? This way we don't have to depend on anything else. 

Buffers already have lots of metadata, I would avoid storing anything extra unless we really have to. The plan for symbol search and goto definition was to just populate a hashtable at parse time.

Mulej: The problem comes with having "/*" and "*/" in strings, that's why I was looking at the color buffer in the first place.
That being said, we could just ignore that edge case for now.

focus-editor: I don't think that's a big problem. We can definitely ignore it

Mulej: Modified to not use the color array. Also just noticed that github ate some asterisks in my previous comment...

focus-editor:
Thanks for your PR!

I've made a couple of minor changes to make the code more like the surrounding code, also added an early return for the default case https://github.com/focus-editor/focus/commit/a6ad5dec7a8e88b1f3c900bb9b6c3671fd058465

PR #83, opened 2023-06-26
Select next word UI

hfr4:
Removed the editor.whole_word variable.
Now it use the searchbar.whole_word / .case_sensitive and you can toggle the options via the existing shortcuts (Alt+W / Alt+C).
There is also new  "minimized" mode for the search bar ui for toggling them.

### DIFF
--- a/FOCUS-CHANGELOG.txt
+++ b/FOCUS-CHANGELOG.txt
@@ -21,9 +21,7 @@
 - [Current file watcher limitation] If you open a file in a large folder, the file watcher might scan the whole folder in order to watch for the changes to the file.
 
 
-# RELEASES ==================================================================
-
-# Version 0.1.11 (UNRELEASED)
+# Main Git Branch
 
 + Linux: The "Open Projects Directory" command as well as shift-clicking files/directories in the "open file" dialogs (Ctrl+O, Ctrl+Shift+O, Ctrl+P) should now work.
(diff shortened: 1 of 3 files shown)


focus-editor:
I've checked out the branch and played with it.

I understand where you're coming from with this change, but I'm not sure I want to merge it, which is a shame because it's a large piece of work.

Reasons:

- The popup which pops up every time you press ctrl-D, while being small, is a bit annoying - it's one step towards the things we're trying to stay away from in this editor, i.e. things jumping out in your face when you haven't asked for them.
- The behaviour when searching selection occurrences is well-established in VSCode (and iirc Sublime and other similar editors) - when you use Ctrl-D to select a word, you get a whole word search, otherwise you don't.  This PR changes that behaviour in a way that people will find surprising.
- Selection search and in-buffer search now share the toggles, which is not what I would personally want. Selection search is a thing that benefits from being transient, you do it quickly and forget, and it won't have any lingering state.

Suggestions:

- Get rid of the popup. 
- Make the flags transient again. When you lose your selections, go back to defaults. Don't share that info with the search bar.
- Keep the Alt-C and Alt-W behaviour. Their effects will be obvious by the change of selection highlights, so it's ok if there's no popup with this extra info.

Thanks for your efforts, your work is appreciated.

focus-editor:
Sorry to see you decide it's not worth it.

I hope there's no hard feelings.

PR #84, opened 2023-06-26
When searching buffers also fuzzy match the whole file path.

Mulej:
Allows you to filter buffers both by path and by filename at the same time. Useful for opening module.jai files for example :)

### DIFF
--- a/src/widgets/open_file_dialog.jai
+++ b/src/widgets/open_file_dialog.jai
@@ -339,7 +339,35 @@ refresh_entries :: (clear_input := true) {
                 name_score, name_highlights := fuzzy_match(buffer_name, filter_chars);
                 path_score, path_highlights := fuzzy_match(buffer.file.path, filter_chars);
                 score := 200 * name_score + path_score;  // name score is much more important
-                if score < 0 continue;
+
+                if score < 0 {
+                    // Try to match the whole path with name.
+                    path_part: [] string;
+                    name_part: [] string;
+
+                    slash_index := -1;
+
+                    for < char, char_index: filter_chars {
+                        if char == "/" {
+                            slash_index = char_index;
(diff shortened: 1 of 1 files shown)


focus-editor:
I've tried this branch but can't see what exactly has changed. We're already searching by the path portion. Please can you give an example of a filter string that works with this change but doesn't without?

Also please note that the filter logic will have to be changed slightly soon, because currently we don't prioritise full matches while we really should

Mulej:
An example would be "Input/module.jai", here I'm specifying both the path and the filename which doesn't work without this change.

> Also please note that the filter logic will have to be changed slightly soon, because currently we don't prioritise full matches while we really should

If the only thing changing is the fuzzy_match internals, then this will still work after the changes. Otherwise it might be a better idea to always match against the full file path (with filename included) and prioritize matches that match the leftmost parts of the path.

valignatev:
I also want this. I can't, for example, ctrp+p for `focus/input.jai` or `focus input.jai` to distinguish from builtin jai modules easily

focus-editor:
Thanks, I see what you mean now.

The reason I'm not merging your PR yet is because I need to take a careful look at the code which does the filtering and maybe redesign it a bit. If after that the design doesn't change then I'll merge your PR. Otherwise I will incorporate it into those changes so that it works.

The original idea was to make it behave more like VSCode and support queries separated by space, so that "jai/modules/Input/module.jai" could be matched by "mod inpu", for example. I just didn't have the time to implement it.

focus-editor:
This has finally had some attention. It should now be possible to search by things separated by space or slash and they will be matched separately against both the name and the path. 

@Mulej if you're still using the editor, I will appreciate feedback on the new fuzzy match logic.

PR #89, opened 2023-06-30
Linux: implement wait_message for X11 and Wayland

ivan-cx:
It's not ideal solution, but I would argue that it's much better solution that existing solution.
Downsides: 
 - up to 10ms delay on keystroke
 - still polling the OS (X11 or Wayland) for events, but with much lower frequency than before
 - no MacOS wait_message implementation

Upsides:
 - don't sleep if redraw_requested (no lags)
 - much less CPU consumption when idle
 - removed platform-dependent delays in main loop (could be bad for MacOS, since it has no wait_message implementation)

### DIFF
--- a/modules/Linux_Display/ldw_input.jai
+++ b/modules/Linux_Display/ldw_input.jai
@@ -253,39 +253,6 @@ pointer_listener :: wl_pointer_listener.{
 };
 
 wl_update_window_events :: (display: *WLDisplay) {
-    display_poll :: (display: *WLDisplay) -> s32 {
-        wlfd := wl_display.get_fd(display.handle);
-
-        pfd: [2]pollfd;
-        pfd[0].fd = wlfd;
-        pfd[0].events = POLLIN;
-        pfd[1].fd = display.epfd;
-        pfd[1].events = POLLIN;
-
-        ret := poll(pfd.data, 2, 5);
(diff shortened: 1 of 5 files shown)


focus-editor:
@ileonte if you like this change please merge.

@cookednick I think if it's merged then some changes might be required from the Mac side as well?

ileonte:
I've tested these changes and unfortunately this doesn't work correctly on Wayland (it breaks key repeat at the very least, possibly other things I haven't run into since I don't have time to test thoroughly). The X11 path seems to work fine at first glance.

We could merge the X11 changes now and look into the Wayland issues later (if you don't have the time/inclination to do it I can look into it sometime later).

PS: It would be helpful if you could split the changes into at least two commits (one for X11 and one for Wayland) so I can cherry-pick only the X11 changes in case we decide to go that route.

ivan-cx:
I think it needs to be re-done completely.
What I would do differently:
Step 1. remove platform-dependent code from the main loop. It's hard to manage now, since all platforms except Windows share the same hack. It should be 3 separate hacks - one for each platform (MacOS, X11, Wayland)
Step 2. fix X11, fix Wayland, fix MacOS - by "fix" I mean "implement WaitMessage"

Alternative route: do not use WaitMessage at all, do it the other way around.

Issue #90, opened 2023-07-03
macOs: Focus doesn't exit fully and prevents shutdown


mzzovko:
**Describe the bug**
MacOs: Focus doesn't exit fully and prevents shutdown when trying to create a new file without a project opened

**To Reproduce** 
On macOS (video is below)
1. Open Focus
2. press `shift + cmd + p` 
3. select `Create new file`
4. type something
5. press `CMD + S`
7. press macOs quit app button

**Expected behavior**
- I expect that CMD + S will save file - but nothing happens or something happens - but it's not clear what 
- I expect that macOs exit button will shutdown app completely
- I expect that Focus doesn't prevent shutdown of my OS

**Screenshots**
https://github.com/focus-editor/focus/assets/9343781/5b05f82f-cc42-429a-8bb6-287f19338973

I feel like this is a symptom of some other underlying problem - how stuff is released from use somewhere

cookednick:
Thanks for the submission. This is a known bug and I intentionally left it sitting around, but it's good to have it documented here.

> I expect that CMD + S will save file - but nothing happens or something happens - but it's not clear what
It currently does nothing. I couldn't find a simple way to bring up the Finder dialog for saving a file from Jai. We also already have plans to write our own cross-platform equivalent. When that's done, saving a new file will work on Mac. Until then, it is a missing feature.

> I expect that macOs exit button will shutdown app completely
I agree. I will look into this.

> I expect that Focus doesn't prevent shutdown of my OS
Are you talking about (A) how when you close the window it can't be quit? Or are you saying when (B) there are unsaved files, you would expect system shutdown to prompt you about the unsaved files?

For the last one, as a workaround, Cmd+Option+Escape is the shortcut to Force Quit apps that cannot be closed. Sorry about that.

focus-editor:
I believe it's somewhat improved, but there's still an issue with closing by clicking on the close button in the titlebar - #470

PR #91, opened 2023-07-04
Only sleep when there's nothing else to do this frame

valignatev:
Even when the window doesn't have focus, let it exhaust the event queue first. At least on X11, when you press the window manager's/DE's "close window" shortcut, it sends DestroyWindow event and FocusOut event right away, so the editor records `.QUIT` event and then sets `input_application_has_focus` to false right away, which prevents `.QUIT` event to be handled this frame and it gets lost, so you can't close the editor unless you press close combination from your config.

Hopefully, this doesn't conflict with the initial purpose of sleeping when the application loses focus

### DIFF
--- a/src/main.jai
+++ b/src/main.jai
@@ -135,15 +135,17 @@ main :: () {
         screen_with_title_bar.h += 200;  // should be enough to include the title bar
 
         #if OS != .WINDOWS {
-            if !Input.input_application_has_focus {
-                // Avoid changing the cursor when the window is not in Focus.
-                // This also fixes the Cmd+Tab cursor-spawning issue where the window still thinks Cmd is held down.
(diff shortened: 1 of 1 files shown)


valignatev:
I know it conflicts with #89 a little bit, but it can be safely merged without much conflicts until that one gets sorted out for wayland

valignatev:
Maybe it has a potential to fix closing issue on macos as well from #90

ileonte:
This looks good to me :+1:

cookednick:
> Maybe it has a potential to fix closing issue on macos as well from #90

Just confirmed that it doesn't, but I appreciate the thinking. That macOS problem only occurs when there are unsaved buffers. It still occurs, so it is unrelated.

PR #97, opened 2023-07-11
Fixed compiler errors introduced by the new arithmetic overflow checks.

nick-celestin-zizic:
Tried compiling with compiler version 0.1.71 and the new arithmetic bounds checks broke some code that relied on the old math bounds check. Also turned on the arithmetic overflow check for debug builds.

### DIFF
--- a/first.jai
+++ b/first.jai
@@ -30,6 +30,7 @@ build :: () {
             options.array_bounds_check = .ON;
             options.null_pointer_check = .ON;
           case "debug";
+            options.arithmetic_overflow_check = .FATAL;
           case;
             compiler_report(tprint("Command-line argument #%, '%', is invalid. Valid options are: 'debug', 'release'.\n", it_index+1, arg));
         }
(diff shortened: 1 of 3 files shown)


first.jai
```diff
@@ -30,6 +30,7 @@ build :: () {
             options.array_bounds_check = .ON;
             options.null_pointer_check = .ON;
           case "debug";
+            options.arithmetic_overflow_check = .FATAL;
```
cookednick, line 33: This PR is awesome, except this line should be inserted right after line 21 rather than here at 33. We assume debug even if "debug" wasn't passed as an argument. The way you did it, just calling `jai first.jai` wouldn't set this flag and therefore the program wouldn't know it's in DEBUG mode.

After making this change, in release, accordingly set a non-`.FATAL` value in this if-statement when the "release" case is true. That should make everything right as rain. I'll test it after these changes and merge. Thank you Zick!

cookednick (approved):

PR #103, opened 2023-07-19
A lot of new C/C++ keywords and directives + more C/C++ file extensions

Roman-Skabin:


focus-editor:
Thanks!

The initial plan was to have separate highlighting for C and C++, but we can always separate that if people complain.

PR #106, opened 2023-07-22
Add crude type parsing for jai tokenizer

rainbowism:
It'll highlights the types in _most_ situations. It isn't a very robust solution but after some testing it seems to work well enough.

### DIFF
--- a/src/langs/jai.jai
+++ b/src/langs/jai.jai
@@ -59,7 +59,12 @@ get_next_token :: (using tokenizer: *Tokenizer) -> Token {
     char := << t;
 
     if is_alpha(char) || char == #char "_" {
-        parse_identifier(tokenizer, *token);
+        if is_type_def(tokenizer) {
+            parse_identifier(tokenizer, *token);
+            token.type = .type_keyword;
+        } else {
+            parse_identifier(tokenizer, *token);
+        }
     } else if is_digit(char) {
         parse_number(tokenizer, *token);
     } else if char == {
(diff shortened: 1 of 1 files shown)


mehlian:
Ok, I'm not participating in the beta, but if I remember correctly, it is possible to ask jai compiler to return data information based on the provided source code. However, from what I can see, this text editor performs source code parsing on its own to identify the tokens. Of course, if this editor is to support several languages, not every compiler can have such functionality.

Have I missed something?

focus-editor:
Thanks for the PR! I tried to do it myself but decided against it because there are too many special cases (see screenshots)

image

image

I'm honestly not sure it's even worth it to try to highlight user types using tokens at this point. The plan was to leverage the compiler eventually to get a list of types asynchronously and put it in a hash map for highlighting (and other things)

rainbowism:
That's fair. I have actually resolved those 2 scenarios but if the plan is for using the compiler rather than a bespoke tokenizer then that seems like a better plan.

focus-editor:
There's more than 2 unfortunately. Also I checked and it slows down the parser by 5-10%, which is not great because we're parsing on every keystroke.

Here's some more (handling these cases would slow it down even further).

image

image

image

Would you agree to wait until we try it with the compiler?

rainbowism:
That's fine with me. I can just use this personally for now in the meanwhile, or somebody can pull in my branch if they wish.

focus-editor:
Great. Thank you for your understanding!

PR #109, opened 2023-07-24
Preload cursors on X11 so you don't have to load them every time

valignatev:
At the moment, every time the editor calls `set_pointer_image`, X11 backend loads and sets the cursor image. It both slows the editor down a lot and makes the cursor flicker constantly. With this PR, we preload X11 cursors during display initialization, when we load libXcursor, and we only set a new cursor when it actually changes, generating way fewer CursorNotify events

### DIFF
--- a/modules/Linux_Display/ldx_display.jai
+++ b/modules/Linux_Display/ldx_display.jai
@@ -22,6 +22,12 @@ X11Window :: struct {
     glx_win: GLXWindow;
 }
 
+x_cursors: Table(string, Cursor);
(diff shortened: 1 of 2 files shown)


ileonte:
Thank you for the fix!

focus-editor:
@ileonte please can you update the changelog with what's been done? Thanks

valignatev:
ah my bad, I'll make sure to update the changelog myself next time

ileonte:
Done.

PR #110, opened 2023-07-27
Add syntax highlighting for GLSL

SentientCoffee:
Very crude at the moment, just copied over the c.jai file and modified it for GLSL keywords and types XD.

To note (see `langs/glsl.jai:922`):
The #inserts to generate KEYWORD_MAP need to be split because trying to generate cases for all 277 (!) keyword values makes the compiler crap itself:

```
Error: Jai encountered an exception running the program and is terminating.

Command line arguments:
   jai.exe first.jai

This happened outside compile-time execution, so it is likely the compiler's fault.

Stack trace of the compiler:

   unknown(C1D87614) : BaseThreadInitThunk
```

When testing, this happens when keywords_count >= 267.
With split #inserts, the compiler can handle them just fine.

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -904,8 +904,12 @@ set_lang_from_path :: (buffer: *Buffer, path: string) {
             case "focus-config";   buffer.lang = .Focus_Config;
             case "cs";             buffer.lang = .CSharp;
 
-            // @Todo: more common extensions?
-            case "glsl";           buffer.lang = .Glsl;
+            case "vert"; #through; // NOTE: Vertex shader extension
+            case "frag"; #through; // NOTE: Fragment shader extension
+            case "geom"; #through; // NOTE: Geometry shader extension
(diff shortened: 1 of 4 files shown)


src/editors.jai
```diff
@@ -904,6 +904,9 @@ set_lang_from_path :: (buffer: *Buffer, path: string) {
             case "focus-config";   buffer.lang = .Focus_Config;
             case "cs";             buffer.lang = .CSharp;
 
+            // @Todo: more common extensions?
+            case "glsl";           buffer.lang = .Glsl;
```
obiwanus, line 908: I think it's a good idea to add those common extensions yes.

obiwanus (approved):
Thanks, that's helpful! Will you please add other extensions such as .vert, .frag, what else?

src/editors.jai
SentientCoffee: I've added `vert`, `frag`, `geom` and `tess`, not sure if there are any more of these that would be helpful?

PR #111, opened 2023-07-29
Add "Save as" and "Delete current file" functionality

SentientCoffee:
Sometimes you want to duplicate a file on your filesystem or just delete it because you goofed. Well now you can do it from the editor without having to switch from your file explorer and back all the time :)

The delete current file option pops up a confirmation dialog just to be sure you want to delete the file, and even if you get past that and decide after that it was an accident, the buffer with all your file data is preserved so you can resave it if necessary.

With both of these functionalities in, a patch for renaming files should be pretty easy, if that's also something we want.

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -421,10 +421,10 @@ recalculate_colors :: (buffer: *Buffer) {
     }
 }
 
-save_buffer_to_disk :: (using buffer: *Buffer, buffer_id: s64) -> saved: bool {
+save_buffer_to_disk :: (using buffer: *Buffer, buffer_id: s64, save_new_file := false) -> saved: bool {
     if readonly return false;
 
-    if !has_file {
+    if save_new_file || !has_file {
         file_path, success := platform_get_save_file_name(get_buffer_name(buffer));
         if success && file_path {
             has_file = true;
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -421,50 +421,88 @@ recalculate_colors :: (buffer: *Buffer) {
     }
 }
 
-save_buffer_to_disk :: (using buffer: *Buffer, buffer_id: s64, save_new_file := false) -> saved: bool {
+save_buffer_to_new_file_on_disk :: (buffer: *Buffer, buffer_id: s64) -> saved: bool {
+    if buffer.readonly return false;
+
+    file_path, success := platform_get_save_file_name(get_buffer_name(buffer));
+    if !(success && file_path) {
+        return false;
+    }
+
+    // Save the file BEFORE making a new buffer so find_or_create_buffer can find it
+    if config.settings.strip_trailing_whitespace_on_save then strip_trailing_whitespace(buffer, buffer_id);
+
+    success = write_entire_file(file_path, to_string(buffer.bytes));
+    if !success then return false;
+
+    new_id, created := find_or_create_buffer(file_path);
+    redraw_requested = true;
+
+    // Reroute all editors using the old id
+    for * editor : open_editors {
+        if editor.buffer_id == buffer_id {
+            editor.buffer_id = new_id;
+            editor.scroll_to_cursor = .yes;
+            for * cursor : editor.cursors { put_cursor_in_valid_spot(cursor, buffer); }
+            organise_cursors(editor);
+        }
+    }
+
+    if !file_is_watched(file_path) then start_watching_file(file_path);
+    update_window_title(open_editors[editors.active].buffer_id);
+
+    return true;
+}
+
+save_buffer_to_disk :: (using buffer: *Buffer, buffer_id: s64) -> saved: bool {
     if readonly return false;
 
-    if save_new_file || !has_file {
+    if !has_file {
+        // Save new file buffer
         file_path, success := platform_get_save_file_name(get_buffer_name(buffer));
-        if success && file_path {
-            has_file = true;
-            file = get_file_info_from_full_path(file_path);
-            set_lang_from_path(buffer, file_path);
-            needs_coloring = true;
-            redraw_requested = true;
+        if !(success && file_path) {
+            return false;
+        }
 
-            lock(*open_buffers_lock);
+        has_file = true;
+        file = get_file_info_from_full_path(file_path);
+        set_lang_from_path(buffer, file_path);
+        needs_coloring = true;
+        redraw_requested = true;
+
+        lock(*open_buffers_lock);
+        {
             existing_buffer_id, found_existing := table_find(*buffers_table, file_path);
-            if found_existing && buffer_id != existing_buffer_id {
-                // The user chose to save the new buffer into an existing file. OK.
-                // Mark the old buffer as deleted and use the new one
-                removed := table_remove(*buffers_table, file_path);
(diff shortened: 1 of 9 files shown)


src/editors.jai
```diff
@@ -1184,22 +1186,24 @@ select_word_or_create_another_cursor :: (using editor: *Editor, buffer: Buffer)
...
 
 save_all :: (editor: *Editor, buffer: *Buffer) {
     new_edit_group(buffer, editor);
 
-    for * buffer, buffer_id : open_buffers {
-        if !buffer.modified continue;
-        save_buffer_to_disk(buffer, buffer_id);
+    for * b, b_id : open_buffers {
+        if !b.modified continue;
+        save_buffer_to_disk(b, b_id);
```
focus-editor, line 1206: Why was this change made?

src/buffer.jai
```diff
@@ -680,7 +742,9 @@ find_or_create_buffer :: (path: string) -> buffer_id: s64, created: bool {
 }
 
 get_buffer_name :: (buffer: Buffer) -> string {
-    if buffer.has_file return buffer.file.name;
+    // NOTE: After deleting using "Delete Current File", `has_file` could be false while buffer.file still exists (because we don't override it).
+    //       So if we fail the `has_file` check, check the `file.full_path` as well to see if there is still a valid name that the buffer can draw.
+    if buffer.has_file || buffer.file.full_path return buffer.file.name;
```
focus-editor, line 747: Please don't do it because if `buffer.has_file` is `false`, the context of `buffer.file` may contain garbage.

src/widgets/delete_file_dialog.jai
```diff
@@ -0,0 +1,45 @@
...
+}
+
+get_buffer_to_delete :: () -> *Buffer {
+    return buffer_to_delete;
+}
+
+#scope_file
+
+editor_to_update: *Editor;
+buffer_to_delete: *Buffer;
```
focus-editor, line 45: This is wildly unsafe. Please never store pointers to editors (or buffers, for that matter), for any prolonged period of time (longer than one function).

Between the time the pointer to editor is stored and used the editor array may be reallocated and we'll crash.

Please use ids instead. Here as it seems you don't need the editor anyway.

focus-editor (changes requested):

src/editors.jai
```diff
@@ -1184,22 +1186,24 @@ select_word_or_create_another_cursor :: (using editor: *Editor, buffer: Buffer)
...
 
-save_buffer :: (editor: *Editor, buffer: *Buffer) {
+save_buffer :: (editor: *Editor, buffer: *Buffer, save_new_file := false) {
     new_edit_group(buffer, editor);
 
     if buffer.has_file {
-        maybe_mark_buffer_as_deleted(editor.buffer_id);  // check in case we missed that it's deleted
-        session_notify_modified_buffer(editor.buffer_id);
+        maybe_mark_buffer_as_deleted(editor.buffer_id);   // check in case we missed that it's deleted
+        session_notify_modified_buffer(editor.buffer_id); // @Cleanup? This doesn't seem to do anything.
```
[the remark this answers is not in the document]
SentientCoffee: Removed the comment.

src/editors.jai
SentientCoffee: I thought it was a possible shadowing bug, I can revert it back if needed.

src/buffer.jai
SentientCoffee: In what scenario would it have garbage?

src/widgets/delete_file_dialog.jai
SentientCoffee: Fixed.

src/buffer.jai
focus-editor: I can't tell you right now, maybe currently there aren't such scenarios, but `has_file` was added there specifically to guard against that.

Maybe we could reconsider and see if we can rely on buffer.file.full_path being empty instead, but I'd need to go and look at the code, which I don't have the time for right now.

I'll try to have a look later and will get back to you on this.

focus-editor:
Thanks for the updated PR. I will have a closer look and test later today when I have some time.

src/buffer.jai
SentientCoffee: This is just a convenience for when you have a buffer that has a file name, from a file that used to exist on disk. If it really is that much of a hassle, I can revert this change as well, but it is nice to have the name already there if all you wanted to do was rename the file with a small tweak to the filename, for example.

src/draw.jai
```diff
@@ -1818,7 +1818,9 @@ draw_delete_file_dialog :: () {
         push_scissor(buffers_rect);
         defer pop_scissor();
 
-        buffer := get_buffer_to_delete();
+        // Not locking here because we're not modifying the buffer
+        buffer_id := get_buffer_id_to_delete();
+        buffer    := open_buffers[buffer_id];
```
focus-editor, line 1823: Why not take a pointer here to avoid a copy? Everywhere else when we get a buffer by id we do it by pointer (except for one place I just found, which should also be changed to a pointer)

src/widgets/delete_file_dialog.jai
```diff
@@ -12,34 +12,40 @@ delete_file_dialog_handle_event :: (event: Input.Event) -> bool {
     return false;
 }
 
-show_delete_file_dialog :: (editor: *Editor, buffer: *Buffer) {
+show_delete_file_dialog :: (editor: *Editor) {
+    // Getting a copy of the buffer here (since we're not modifying it) so no need to lock
+    buffer := open_buffers[editor.buffer_id];
```
focus-editor, line 17: Same here

src/widgets/delete_file_dialog.jai
```diff
@@ -12,34 +12,40 @@ delete_file_dialog_handle_event :: (event: Input.Event) -> bool {
...
 
 close_delete_file_dialog :: () {
-    editor_to_update = null;
-    buffer_to_delete = null;
     activate_editors();
 }
 
 delete_buffer_file_and_update_editor :: () {
+    lock(*open_buffers_lock);  // Is this necessary here?
+    defer unlock(*open_buffers_lock);
```
focus-editor, line 30: We only need to lock when we're adding a new buffer to the buffer array or are modifying the path-to-id hash table.

We're not doing it here, and the `delete_buffer_from_disk` function, which does it, locks the corresponding portion of the code itself, so it should be safe to not lock here in this function.

focus-editor (changes requested):
Just a couple of small things. 

I'm ready to merge when those are addressed. Regarding the `has_file` thing, let's merge it as is and I will make a note for myself to review the use of `has_file`.

Thanks

src/draw.jai
SentientCoffee: Changed to a pointer.

src/widgets/delete_file_dialog.jai
SentientCoffee: Changed to a pointer.

src/widgets/delete_file_dialog.jai
SentientCoffee: Removed.

focus-editor:
Thank you @SentientCoffee !

PR #112, opened 2023-07-30
Rewrite X11 events polling. Implement platform_wait_message for X11

valignatev:
This change makes an editor consume no CPU while idle. Here's what changed:
1) Implemented platform_wait_message for Linux, X11 specifically. It needs a window for X11 display file descriptor, so I took the liberty to change platform_wait_message interface, now it takes Window_Type argument. Windows doesn't use it, macos version of the call isn't implemented yet.

2) X11's platform_wait_message uses select(2) to wait on Display's file descriptor being ready for reading and then confirms with XPending that it's actually a user-level event and not some internal xlib business.

3) XNextEvent is replaced with XCheckIfEvent. XCheckIfEvent doesn't hold a global xlib mutes, it retrieves an event in an atomic and non-blocking way. It makes our event loop thread-safe, and we don't need to surround event handling with XLock/XUnlockDisplay anymore.

4) Pulled a few sys/select.h-related things into ldx_input.jai

5) Reduced CPU usage hack that's in the main loop is only working on MACOS now. It means that Wayland-based CPU usage might slightly go up (although probably not by as much as it was before this hack got introduced, due to earlier cursor setting optimizations), but I hope that we'll just implement Wayland-based platform_wait_message since all the plumbing is already in-place. Unfortunately, I don't yet have a working Wayland compositor that I would use all the time to implement it myself and test properly
image

### DIFF
--- a/modules/Linux_Display/ldx_input.jai
+++ b/modules/Linux_Display/ldx_input.jai
@@ -75,8 +75,8 @@ x11_translate_key_code :: (key: KeySym) -> Key_Code {
     return UNKNOWN;
 }
 
-x11_wait_for_events :: (window: *X11Window) {
-    x11_display := window.display;
+x11_wait_for_events :: () {
+    x11_display := isa(global_display, X11Display);
 
     xev: XEvent;
     needs_flush := true;
--- a/modules/Linux_Display/ldx_input.jai
+++ b/modules/Linux_Display/ldx_input.jai
@@ -75,17 +75,81 @@ x11_translate_key_code :: (key: KeySym) -> Key_Code {
     return UNKNOWN;
 }
 
-x11_update_window_events :: (x11_display: *X11Display) {
-    display := x11_display.handle;
-    XLockDisplay(display);
+x11_wait_for_events :: (window: *X11Window) {
+    x11_display := window.display;
 
     xev: XEvent;
-    while XPending(display) {
-        XNextEvent(display, *xev);
-        x11_handle_event(display, *xev);
-    }
+    needs_flush := true;
+    while true {
+        if needs_flush XFlush(x11_display.handle);
 
-    XUnlockDisplay(display);
+        // Example from select(2) and most reference code uses FD_ZERO macro, but jai zeroes structs out by default.
+        fds: fd_set;
+        FD_SET(x11_display.file_descriptor, *fds);
+        result := select(
+            x11_display.file_descriptor + 1,
+            *fds, // file descriptors for read
+            null, // file descriptors for write
+            null, // file descriptors for error detection
+            null, // timeout setup. null means wait indefinitely
+        );
+
+        if result < 0 {
+            error := errno();
+            if error == EINTR {
+                // According to man 7 signal, select always fails with EINTR if
+                // it gets interrupted by a signal, even if then the process gets
+                // restarted, so we just keep going. EINTR can also happen if something
+                // wrong with the timer, but we're not using it
+                continue;
+            }
+            // Returning on other errors. If one of these happens the application
+            // is most likely to crash anyway
+            if error == EBADF || error == EINVAL {
+                log_error(
+                    "Invalid X11 file descriptor on X11 display. Error code is %. See select(2)\n",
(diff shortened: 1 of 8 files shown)


modules/Linux_Display/ldx_input.jai
```diff
@@ -75,17 +75,81 @@ x11_translate_key_code :: (key: KeySym) -> Key_Code {
...
+
+    xev: XEvent;
+    needs_flush := true;
+    while true {
+        if needs_flush XFlush(x11_display.handle);
+
+        // Example from select(2) and most reference code uses FD_ZERO macro, but jai zeroes structs out by default.
+        fds: fd_set;
+        FD_SET(x11_display.file_descriptor, *fds);
+        result := select(
```
ileonte, line 89: Could we use `poll()` here like we do for Wayland? ref: https://github.com/focus-editor/focus/blob/main/modules/Linux_Display/ldw_input.jai#L259

modules/Linux_Display/module.jai
```diff
@@ -365,6 +365,13 @@ timer_destroy :: (timer: Timer) {
     <<timer = .{};
 }
 
+wait_for_events :: (window: LDWindow) {
```
ileonte, line 368: Is there a reason this needs the window handle here? Unless I'm missing something `x11_wait_for_events()` only needs the display and we already have that in `global_display`.

src/platform/windows.jai
```diff
@@ -84,7 +84,7 @@ platform_set_window_title :: (title: string) {
     SetWindowTextW(window, utf8_to_wide(title));
 }
 
-platform_wait_message :: inline () {
+platform_wait_message :: inline (window: Window_Type) {
```
ileonte, line 87: I'm guessing this is maybe a left-over change you accidentally included.

modules/Linux_Display/ldx_input.jai
valignatev: I went with select because winit folks had some trouble with poll, and glfw uses select as well. But SDL, on the other hand, uses poll. I'll try it out

modules/Linux_Display/module.jai
valignatev: I intentionally changed the interface to take window for all platforms. I felt like bundling a pointer to a display and its file descriptor together is handy as opposed to just shove the display file discriptor globally near it and then manage two separate globals. But I'll change it if you think that it's better this way

src/platform/windows.jai
valignatev: see above

modules/Linux_Display/module.jai
ileonte: > I felt like bundling a pointer to a display and its file descriptor together is handy as opposed to just shove the display file discriptor globally near it and then manage two separate globals
I'm not sure what you mean by this. You already added the filedescriptor as a member of `X11Display`, so they are already bundled together under `global_display`.

Anyway, the idea is that this API is intended to work with multiple windows (it's called from `Window_Creation`/`Simp` among other things, which support multiple windows), so unless it explicitly needs a specific window for this (which I don't think it does by looking at `x11_wait_for_events()`) it should just use `global_display` to get to the handle.

src/platform/windows.jai
ileonte: Right, I missed the MacOS change. Still, as I said above, if all you need if the display handle `global_display` is the way to go, all of this should work with multiple windows anyway.

modules/Linux_Display/module.jai
valignatev: ahh I see what you mean, for some reason I thought that global_display is `*Display` (from xlib), not `*X11Display`. I'll change it then

valignatev: Poll seems working!

modules/Linux_Display/ldx_input.jai
valignatev: Poll seems working

modules/Linux_Display/module.jai
valignatev: Used global_display, reverted the interface change so platform_wait_message doesn't accept any args as it used to

PR #114, opened 2023-08-01
Request to redraw when clicking on finder result with a mouse

valignatev:
Otherwise, an event loop will block and wait for the next event without actually redwaring the window with the result shown. File finder already does that


obiwanus:
Danke schön

Issue #66, opened 2023-06-14
Unable to save file


Quertin:
Unable to save file created with ctrl-n by pressing ctrl-s or the save dialog when quitting focus.

Focus only outputs "name = 'filename" to console when pressing ctrl-s and no errors,
the "save" button in the quit focus dialog does nothing only "cancel" and "don't save" do anything.
I also didn't find a way to determine which folder it tries to write to.

I tried default config file and running focus as sudo.

Linux Mint 21.1 Cinnamon with 5.19.0-43-generic Kernel.
focus-linux Version 0.1.10.

![b](https://github.com/focus-editor/focus/assets/136612013/4a6e4097-0093-46f3-85c5-3b71066e85c3)

focus-editor:
On Linux unfortunately this is not implemented yet. There will be a custom file save dialog since there isn't any standard file explorer like on Windows or macOS that we could use.

ivan-cx:
I'd like to look into that, maybe implement some very basic save dialog, if there are no objections :-)

focus-editor:
@ivan-cx sure if you want to take a stab, please do. But if you want it merged it may require some back and forth to make sure the UI is in alignment with the editor's philosophy. Just saying because apparently some people were upset when their pull requests weren't merged immediately.

ivan-cx:
@focus-editor thanks, good to know! That's why I want to start with something minimal :-)

ivan-cx:
@focus-editor so, I got stuck on this (and I think it would be useful for other implementation attempts in the future): 

on Windows you use **GetOpenFileNameW** inside **platform_get_save_file_name**, which (if I understand correctly) blocks the main loop and render, and returns the target path "immediately" to our program. Please correct me, if I'm mistaken here.

So, with such architecture we would be forced to use some blocking OS-specific UI (since our main loop is blocked and we can't render our own UI), which... could be an option, but who would guarantee that each OS has such special UI? 

I would try to implement single solution for all platforms with an option to rollback to original way on Windows. I'm not expecting to merge this experiment to the mainline, but maybe it would be useful for future attempts to solve this.

focus-editor:
Yes it's not a trivial piece of work. You'd need to implement another widget similar to the open file dialog and activate it instead of calling the standard platform dialog. 

We'd still like to keep the possibility to use the standard dialog on Windows or macOS because some users prefer that, so we'd need a boolean flag in the config - something like `use_custom_file_save_dialog: true`

[...]

PR #118, opened 2023-08-03
Sketch out built-in Save File Dialog (#66)

rexim:
Related to #66

This is the beginning of the built-in Save File Dialog implementation. Since implementing it properly is a relatively significant task (especially for somebody like myself who's not familiar with the inner workings of the Editor that well), this commit simply outlines how it might potentially work.

What's needs to be done

1. Cool Open File Dialog Style navigation. Save File Dialog is just a single input and you have to type out the full path to the file you want to save
2. I took the save logic from `save_buffer_to_new_file_on_disk()` and `save_buffer_to_disk()` and tried to replicate it in an asynchronous fashion in `save_file_dialog_handle_event()`. It seems to work alright for me. The occurred code duplication between `save_buffer_*()` and `save_file_dialog_handle_event()` needs to be compressed properly somehow, otherwise it may create problems in the future when the logic needs to be updated. We don't want to update it in 2 places separately.
3. Configuration flag that controls whether to use the built-in Save File Dialog or the Native OS one.

Please, let me know if there are any problems with the Pull Request or if you want me to make any changes.

Thanks!

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -61,6 +61,7 @@ draw_frame :: () {
 
     if active_global_widget == {
         case .open_file_dialog;         draw_open_file_dialog();
+        case .save_file_dialog;         draw_save_file_dialog();
         case .delete_file_dialog;       draw_delete_file_dialog();
         case .finder;                   draw_finder();
         case .go_to_line_dialog;        draw_go_to_line_dialog(active_pane_rect);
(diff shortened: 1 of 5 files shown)


focus-editor:
Thanks. Anything is better than what was there before :)

I'll add a note to revisit this code and fix the issues.

PR #119, opened 2023-08-04
added in support for save_on_edit flag

WWilliams741:


focus-editor:
I'm sorry, but saving on every keystroke is rather expensive for large files, and while it works, we wouldn't want to advertise it as a valid option supported by the editor. You're encouraged to use it in your own fork though if it works for you, a few people have already started supporting their forks.

A better way to achieve the same result would be to use memory-mapped files, but that's OS-specific and we probably aren't ready to accept a large PR adding this functionality at this stage.

WWilliams741:
Thanks for the speedy reply. I figured something like this was wrong with it, but I don't know enough about the codebase to make a larger more performant change. Just do the straightforward thing in the least amount of changes. However, the fact that I could make this kind of change in only a couple lines of code speaks volumes to how simple and straightforward this editor is. I couldn't imagine making a change like this on some other editor like VSCode. Keep up the good work.

While I'm here:
Is there a timeline for these kinds of features/plans somewhere? Or are you guys playing it by ear on an as needed basis?

PR #123, opened 2023-08-07
New feature: multi cursor shrink towards origin

WWilliams741:
This works, but I think someone needs to check over my work. I'm a little unfamiliar with this codebase so I mainly copied the structure of create_cursor then modified it from there based on what I saw with some print outputs. Logic is based around main_cursor and original_cursor positions.

This way of creating new cursors makes more sense to me because if you mess up and go one past where you want to you don't have to start over and can instead shrink and/or move the cursors then keep growing and is therefore more versatile. This brings it more in line with the way highlighting lines works (holding shift then arrow keys), which is go in one direction then shrink back towards the cursor if you start going in the opposite direction afterwards.

This brings the editor more in line of what the [(10x editor)](https://10xeditor.com/) does by default, except the user can toggle between the two using the config flag. So, whichever way they'd like.

Also, I think you know this already but I kept randomly getting free_list_pop seg faults with backtraces. Made it a little hard to debug my code, but I think you are in the middle of fixing this right now and are aware of the issue.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,13 @@
+# Contributing
+
+The base editor is intended to be kept simple forever. It means it will likely be "complete" at some point. It also means it will not include very many features.
+
+*Pull requests containing features are more likely to be rejected*. However, you are encouraged to fork the editor and implement anything you need on top. 
+
+We welcome bug reports and bug fixes.
+
+The bigger the change, the more likely it will be rejected. Please don't work on big changes that you want merged before checking with us on Discord or on GitHub.
+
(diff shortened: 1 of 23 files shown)


focus-editor:
Thanks for the PR. I would like to focus on one thing at a time so I will get back to this PR later, but the feature itself seems reasonable. I might ask to change the code a little bit though. Also the name of the option might be a bit confusing. Perhaps we should just make this behaviour default.

focus-editor:
Thanks for your PR.  

I will rewrite the code because it's easier to do that than to explain what I'd like to see

PR #124, opened 2023-08-07
More x11 im hanling

valignatev:
This fixes a few potential segfaults related to X11 input methods and generally makes detecting locales way more robust.
We can still go further - explicitly support ibus and fcitx (with current method, they are supported indirectly with fallback to `"@im=none"`, but _technically_ it's not 100% correct) and also check applicable input styles before creating input context. But this is a good 90% solution

### DIFF
--- a/modules/Linux_Display/ldx_display.jai
+++ b/modules/Linux_Display/ldx_display.jai
@@ -44,13 +44,14 @@ x11_init_display :: (display: LDDisplay) -> bool {
 
     // Default locale, which is "C", doesn't support multibyte characters,
     // so we do this to get system locale which is hopefully set up correctly
-    // result will either by a correct locale from the environment,
+
+    // Result of setlocale will either be a correct locale from the environment,
     // or NULL if setlocale fails, or it will be "C" if somebody exported LC_ALL=""
     new_locale := to_string(setlocale(LC_ALL, ""));
 
     // We need to check whether current locale is supported by X server. It's not
     // guaranteed regardless of whether setlocale call is successful or not.
-    // If setting locale didn't work, we try our best to get it from few well-known
+    // If setting locale didn't work, we try our best to get it from a few well-known
     // environment variables in case one of them is set up correctly. glibc's setlocale
     // tries $LC_ALL first, then a category you specify in the first argument to setlocale,
     // and then $LANG env variable. The issue here is that glibc bails out of this sequence
@@ -62,9 +63,9 @@ x11_init_display :: (display: LDDisplay) -> bool {
         // getenv can have two "falsy" outcomes - env variable exists but is an empty string
         // and env variable doesn't exist. In first case we get a pointer to an empty string,
         // in the second case we get NULL as a result. Converting to jai string covers both cases
-        lc_ctype := to_string(getenv("LC_CTYPE"));
+        lc_ctype := trim(to_string(getenv("LC_CTYPE")));
         if !lc_ctype {
-            lang := to_string(getenv("LANG"));
+            lang := trim(to_string(getenv("LANG")));
 
             if !lang {
                 log_error("LANG environment variable is not found, trying to fall back to C.UTF-8\n");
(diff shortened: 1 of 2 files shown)


modules/Linux_Display/ldx_display.jai
```diff
@@ -30,35 +30,81 @@ preloaded_x_cursors :: string.[
...
-    setlocale(LC_ALL, "");
+    // Default locale, which is "C", doesn't support multibyte characters,
+    // so we do this to get system locale which is hopefully set up correctly
+    // result will either by a correct locale from the environment,
+    // or NULL if setlocale fails, or it will be "C" if somebody exported LC_ALL=""
+    new_locale := to_string(setlocale(LC_ALL, ""));
+
+    // We need to check whether current locale is supported by X server. It's not
+    // guaranteed regardless of whether setlocale call is successful or not.
+    // If setting locale didn't work, we try our best to get it from few well-known
```
ileonte, line 53: Typo: `from few` -> `from a few`.

modules/Linux_Display/ldx_display.jai
```diff
@@ -30,35 +30,81 @@ preloaded_x_cursors :: string.[
...
+    // and then $LANG env variable. The issue here is that glibc bails out of this sequence
+    // as soon as it sees that something is set, even if it's set incorrectly. So the call to
+    // setlocale will fail. That's why we try this process manually.
+    // This process vaguely resembles the way musl gets current locale and looks reasonable enough
+    if !new_locale || new_locale == "C" || !XSupportsLocale() {
+
+        // getenv can have two "falsy" outcomes - env variable exists but is an empty string
+        // and env variable doesn't exist. In first case we get a pointer to an empty string,
+        // in the second case we get NULL as a result. Converting to jai string covers both cases
+        lc_ctype := to_string(getenv("LC_CTYPE"));
```
ileonte, line 65: I think we also `trim()` here, just in case someone has just spaces in `LC_TYPE`.
Also - we should probably do this for all `getenv()` results.

modules/Linux_Display/ldx_display.jai
```diff
@@ -30,35 +30,81 @@ preloaded_x_cursors :: string.[
...
+                new_locale = to_string(setlocale(LC_ALL, "C.UTF-8"));
+            } else {
+                new_locale = to_string(setlocale(LC_ALL, temp_c_string(lang)));
+            }
+        } else {
+            new_locale = to_string(setlocale(LC_ALL, temp_c_string(lc_ctype)));
+        }
+
+        if !new_locale || !XSupportsLocale() {
+            error :: #string __error
```
ileonte, line 80: Not related to this PR, mostly just venting - we need to find a way to get these messages into the log file, we can't really expect everybody to run Focus in a terminal and they would be helpful when troubleshooting.

valignatev: true, I agree

modules/Linux_Display/ldx_display.jai
```diff
@@ -30,35 +30,81 @@ preloaded_x_cursors :: string.[
...
+                log_error("LANG environment variable is not found, trying to fall back to C.UTF-8\n");
+                new_locale = to_string(setlocale(LC_ALL, "C.UTF-8"));
+            } else {
+                new_locale = to_string(setlocale(LC_ALL, temp_c_string(lang)));
+            }
+        } else {
+            new_locale = to_string(setlocale(LC_ALL, temp_c_string(lc_ctype)));
+        }
+
+        if !new_locale || !XSupportsLocale() {
```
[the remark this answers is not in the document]
valignatev: I don't think we do. I've read that XSupportsLocale isn't guaranteed to succeed even if setting the locale does. I wasn't able to somehow trigger it myself, but that's what people on the internet say: https://github.com/rust-windowing/winit/pull/1445
It is probably unlikely that this call actually happens twice

modules/Linux_Display/ldx_display.jai
valignatev: trimmed getenvs in this PR, we can trim other few of them in the project in a separate PR so we don't conflate this one

PR #126, opened 2023-08-08
Don't restart the animation if it's already going

valignatev:


src/utils.jai
```diff
@@ -322,6 +322,15 @@ start_animation :: (anim: *Tween_Animation(T), start: $T, target: T, speed: Time
...
     if speed > 0 then anim.speed = speed;
+
+    // Don't restart the animation at the current frame if it's already started.
+    // But still cap it's starting time at just a single frame delta to be conservative.
+    // This lets us start animation smoothly, as in the "else" branch below, but
+    // continue in a snappy way as we scroll.
+    if anim.started_at {
+        anim.started_at = frame_time - MAX_FRAME_DT;
+        return;
+    }
```
focus-editor, line 333: We never reset `started_at` for any animation when it's done, so essentially what you've done here is forced all animations to become snappy.

This is a problem because it reduces the smoothness of certain animations, e.g. when scrolling by moving the cursor to the edge of the screen - the animations become more jerky.

I'll be happy to accept it if you just make all animations for Linux snappy, since obviously it fixes a much bigger problem on Linux than it creates.

We will still need to revisit the scrolling code and make it work properly in the future.

valignatev: it doesn't make all animations snappy by default, only on subsecuent frames when it's already going. At least it feels different from just having animation snappy from the start. I can reset the animation after it's done though. I hope that's the right place to do it

valignatev: But if you think it's not the right way to do it I can just force snappy scrolling animation for linux, even though it feels a bit different

focus-editor: It does make all animations snappy after the first time the animation is started. I've tested it on 2 Focus instances running side by side.

I don't think resetting the scroll animation's `started_at` is a good idea, because it's just one of the animations, and the change affects all the animations. 

If you want you could add that block for Linux only - I'll be happy with that for now.

valignatev: Ok-ok, did that

focus-editor:
@ileonte does this change fix the problem with an unlocked mouse wheel that you mentioned the other day?

Sorry can't test it myself because my mouse is embarrassingly plain and simple, for the common people :)

ileonte:
Yes, it does. But that also "worked" when I disabled smooth scrolling. I can't tell if what this change is doing is basically the same thing or not.

focus-editor:
Disabling smooth scrolling just disables scroll animations (they jump instantly instead of interpolating), so those are different things

ileonte:
I still get "micro" stuttering when I use the keyboard shortcuts to scroll. But spamming the scroll-wheel no longer completely freezes the animation.

PR #127, opened 2023-08-09
Add option to disable curosr blinking

valignatev:
cursors_stop_blinking doesn't have a config check in case someone disables blinking with live-reloading. So we can successfully kill the refresh timer and set cursors_blinking to false


focus-editor:
Thanks for your PR.

I had thought about making the curosr blink time configurable, but never got around to it. That way we could allow people to adjust the time their curosrs blink for, as well as reducing it to 0, essentially disabling blinking completely. Some people really love to see their curosrs all the time!

I hope you're ok if I add that option and remove the binary flag.

valignatev:
yeah sure, as long as you can disable it, it doesn't matter if it's a binary flag or a scale

focus-editor:
FYI https://github.com/focus-editor/focus/commit/d0c958a05f9774c0c371d3772a600f486819f700

PR #128, opened 2023-08-10
Fix search bar dropdown crashing when there are no results in the list

SentientCoffee:
If you hit the `switch_to_left_editor` keybind while in the dropdown search bar while the dropdown search bar has no results, the editor will fail on an array bounds check. This simple fix makes that not happen.


focus-editor:
Thanks. I couldn't reproduce it, because when a dropdown is active the keybind won't do anything - not sure how you managed to trigger it

SentientCoffee:
Note it's not the "search entire project" dropdown, it's the local editor search bar in dropdown mode (the one bound to Alt-F by default).

focus-editor:
Yes I tested it with that one

PR #130, opened 2023-08-11
add windows window border color effect

amasondev:
Screenshot 2023-08-10 225615

This makes the editor window look a lot cleaner on Windows. Requires bringing in DwmSetWindowAttribute from Dwmapi, which is not yet defined in jai's Windows module. Stubbed in noop procedures for the other platforms. I can take a look at a macOS implementation too. MacVim has it.

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -57,6 +57,9 @@ copy_whole_line_without_selection:      false
 editor_history_size:                    1024
 line_wrap_is_on_by_default:             false
 show_line_numbers:                      false
+dark_titlebar:                          true
+# Windows 11 compatible only
+colored_titlebar:                       true
 
 
 [[keymap]]
--- a/src/config.jai
+++ b/src/config.jai
@@ -342,6 +342,8 @@ Settings :: struct {
     editor_history_size                 := 128;
     line_wrap_is_on_by_default          := false;
     show_line_numbers                   := false;
+    dark_titlebar                       := true;
+    colored_titlebar                    := true;
 
     // TODO
     // convert_tabs_to_spaces_on_load:     false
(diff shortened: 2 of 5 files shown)


focus-editor:
OMG it looks awesome.

If only you'd sent it 30 minutes earlier, it would be a perfect addition to the 0.2.0 release.

I'll test it

focus-editor:
image

Doesn't seem to work for me. What could be the reason?

amasondev:
> image
> 
> Doesn't seem to work for me. What could be the reason?

https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute

> DWMWA_CAPTION_COLOR
> Use with [DwmSetWindowAttribute](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute). Specifies the color of the caption. The pvAttribute parameter points to a value of type [COLORREF](https://learn.microsoft.com/en-us/windows/win32/gdi/colorref).
> 
> Specifying DWMWA_COLOR_DEFAULT (value 0xFFFFFFFF) for the color will reset the window back to using the system's default behavior for the caption color.
> 
> This value is supported starting with Windows 11 Build 22000.

Seems to only be supported on Windows >11

amasondev:
It's possible to at least use a dark title bar on 10 using DWMWA_USE_IMMERSIVE_DARK_MODE, which needs to have a value of 19 on some old versions and 20 on newer versions 🤷

focus-editor:
Tested on Windows 11, looks amazing.

If you add that dark title bar option we could include that as well

amasondev:
Added the dark mode setting, and added config options for both of these. I also removed the setting of the border color, as it tended to lose the contrast against other windows.

focus-editor:
Thanks. I will merge and rename the options to make them a bit more descriptive if you don't mind.

Also, the dark titlebar looks really terrible and has a bug when you just start the editor (resizing the window fixes it)
image

I think it's best if we don't make it default

amasondev:
Oof, yeah sorry I wish I had a W10 VM handy for testing.

PR #131, opened 2023-08-11
Hide cursor when typing

amasondev:


src/platform/windows.jai
```diff
@@ -247,8 +247,20 @@ platform_set_border_color :: () {
...
 }
 
+platform_show_cursor :: (show: bool) {
+    if show {
+        if cursor_shown <= 0
+            cursor_shown = ShowCursor(1);
+    } else {
+        while cursor_shown >= 0 {
+            cursor_shown = ShowCursor(0);
+        }
```
focus-editor, line 257: Why do we have to do it in a loop?

amasondev: > https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showcursor
> ## Parameters
>
> [in] bShow
> 
> Type: BOOL
> 
> If bShow is TRUE, the display count is incremented by one. If bShow is FALSE, the display count is decremented by one.
> 
> ## Return value
> Type: int
> 
> The return value specifies the new display counter.

ShowCursor increments a counter, so to hide the cursor you need to call it until that counter is 0.

focus-editor:
Thanks, looking good!

PR #132, opened 2023-08-12
Colored title bar on macOS

amasondev:
Brings in Objective_C module (Version: beta 0.1.073, 10 August 2023) with some additions. Makes dark titlebar default on macOS due to the title text being black otherwise.


focus-editor:
As discussed on Discord, let's keep this open in case we have to pull in the objective C modules for something other than title bar color, then we can add the required change.

cookednick:
This is great. If you haven't already, you can submit your modified ObjC module to rluba in the SB Discord and he will probably accept changes to the official jai module. Then we could merge a simpler version of this PR.

PR #138, opened 2023-08-19
New « align_cursors » command

hfr4:
New command to align multiple cursors.
default shortcut Alt+A.

https://github.com/focus-editor/focus/assets/30574115/65fd7f24-43e4-4789-ad24-3b550ec6aab1


focus-editor:
That's very useful, thanks.

I haven't had the time to test any corner cases, but given that it's all limited to one command it should be OK for now.

Another thing - it should be possible to avoid rescanning for each cursor, like it's done in other places. You can just cache the coords in a temporary array, or do something similar. I'll have another look at the code later.

PR #143, opened 2023-08-22
new select_all_occurences command

hfr4:
Tried to follow how this work on VSCode.
Maybe the behavior is not optimal especially when using the command with more than one cursor on the editor.

Feedback appreciated.


focus-editor:
Thanks, I will have a closer look later when I have a moment - been busy at work. I had a quick look and I feel like it could be done simpler, but need to spend some more time on it.

PR #149, opened 2023-08-24
add toggle_fullscreen command

ustance:
Since there is currently no way to go into fullscreen on Windows.

### DIFF
--- a/src/platform/linux.jai
+++ b/src/platform/linux.jai
@@ -197,6 +197,11 @@ platform_create_window :: () {
     }
 }
 
+platform_toggle_fullscreen :: () {
+    fullscreen = !fullscreen;
+    toggle_fullscreen(window, fullscreen, *fullscreen_info);
+}
+
 platform_maximize_window :: (window: Window_Type) {
     LD.maximize_window(window);
 }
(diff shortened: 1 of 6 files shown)


src/platform/linux.jai
```diff
@@ -197,6 +197,11 @@ platform_create_window :: () {
     }
 }
 
+platform_toggle_fullscreen :: () {
+    fullscreen = !fullscreen;
+    toggle_fullscreen(window, fullscreen, *fullscreen_info);
```
focus-editor, line 202: I'm assuming you meant `fullscreen_state` here and in `macos.jai`

src/platform/macos.jai
```diff
@@ -118,6 +123,9 @@ platform_show_cursor :: (show: bool) {
 
 #scope_file
 
+fullscreen: bool;
+fullscreen_state: Saved_Window_Info;
```
focus-editor, line 127: Does it work if this is put in `main.jai`? If yes then I'd propose to move it there to avoid unnecessary duplication.

focus-editor (changes requested):

src/platform/macos.jai
ustance: done

PR #156, opened 2023-08-29
Highlight line with cursor

SentientCoffee:
Highlight the line(s) with cursors on them, since without this the cursor can be lost easily if you don't know where it is. Configurable within the config file.

NOTE: DO NOT MERGE THIS YET!!
As of 29 August 2023, there is a compiler bug that doesn't allow me to add to the Settings struct, something to do with compile time code generation. Once that is fixed I will update this to _actually_ be configurable in the config file.

### DIFF
--- a/default.focus-config
+++ b/default.focus-config
@@ -59,6 +59,7 @@ line_wrap_is_on_by_default:             false
 show_line_numbers:                      false
 colored_titlebar:                       true   # Windows 11+ only
 hide_mouse_when_typing:                 true
+highlight_line_with_cursor:             false
 
 [[keymap]]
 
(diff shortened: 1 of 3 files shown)


SentientCoffee:
By changing the `KEYWORD_MAP` size in `src/langs/jai.jai` from `2 * (KEYWORDS.count + TYPE_KEYWORDS.count + VALUE_KEYWORDS.count)` to `3 * (KEYWORDS.count + TYPE_KEYWORDS.count + VALUE_KEYWORDS.count)`, I could finally test out the config changes as well. This PR should be good to merge, although not sure what to do to address being able to build it without this hack.

PR #158, opened 2023-09-02
Updated toggle_comment to support tabs, and be more robust

Sanian-Creations:
Try it out, comment some lines, set your settings to spaces, to tabs, place multiple cursors in random places. See if it does anything weird. Your code should look exactly the same after commenting and un-commenting it. I'm pretty sure I've covered my bases.

## Example
Before, it could mess with your code layout.
Take this example. Every `|` is a cursor, not a character.
```
some text
|



other text
|
```
Running toggle_comment would turn that, into this:
```
some text
    // other text

```
Which both comments out a line that it shouldn't have, and removed a bunch of lines in between.
Now, it just does nothing because it correctly realizes both cursors are placed on empty lines.

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -2733,60 +2733,6 @@ toggle_comment :: (editor: *Editor, buffer: *Buffer) {
         }
     }
 
-    whitespace_get_visual_width :: inline (whitespace: string, tab_size: int) -> int {
-        // Given a string of whitespace, will return the visual width of said whitespace.
-        // Example: tab_size=4, "\t  \t " has 5 characters but due to tab alignment it looks like it's 9 wide.
-        indent := 0;
-        for cast([]u8) whitespace { indent = add_visual_indent(indent, it, tab_size); }
-        return indent;
-    }
-
-    whitespace_count_chars_in_visual_width :: inline (whitespace: string, visual_width: int, tab_size: int) -> int {
-        // Given whitespace and a visual width, will count the number of characters that fill that visual space.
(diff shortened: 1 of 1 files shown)


src/editors.jai
```diff
@@ -2724,88 +2724,186 @@ paste_from_clipboard :: (editor: *Editor, buffer: *Buffer) {
...
+    add_comment             := false;
+
+    for 0..last_line {
+        macro_every_loop(it); //vars: is_empty_line, start, str, whitespace, trimmed_str
+        if is_empty_line continue;
+
+        whitespace_visual_width := whitespace_get_visual_width(whitespace, tab_size);
+
+        if min_indent_visual_width < 0 || whitespace_visual_width < min_indent_visual_width
+            then min_indent_visual_width = whitespace_visual_width;
```
focus-editor, line 2835: One thing I would really appreciate is if those who sent PRs respected the surrounding code's style a bit more. I'm not a code style warrior but I do value consistency. In this codebase we never put an if block on a new line without braces, for instance. There are also other small things, which makes it a bit frustrating to review.

Sorry I don't have the time to finish the review yet as this is a lot of code. I feel like it could be done with fewer changes, but will need to spend more time on it later.

Sanian-Creations: I've changed the style up a little to match the rest of the code better, I hope.
One thing that's still different is I defined some procs inside the body, I'm not really sure where else to put them.
If you've got more comments on the implementation when you do have the time, I'd be happy to discuss them here or on the discord.

focus-editor:
Thanks, works as expected 👍

PR #161, opened 2023-09-03
Dont free intividual files on non windows

valignatev:


modules/Input/module.jai
```diff
@@ -221,11 +221,6 @@ input_per_frame_event_and_flag_update :: () {
     // Called once per frame, probably.
 
     for events_this_frame {
-        #if OS != .WINDOWS {
```
ileonte, line 224: We can't do this, it would introduce a memory leak on MacOS. We should really just stop allocating the paths with the temp allocator on Linux to match what the other platforms do.

ileonte: (just for clarity - my comment is about this entire block that's being removed not this one single line)

ileonte: Personally I almost never use drag-and-drop for anything ever, but who knows how people are used to doing things.

Anyway, I think we should change the `Input` module to buffer the paths "per-frame" (similar to how it buffers input events), allocating the strings with the `temp` allocator. Then, a special `get_dropped_files` function would `copy_string` those temp-allocated strings into a caller specified array which the use can then free whenever they are done with those paths (the module then reserves 'the right' to reset its internal paths array every frame).

valignatev: I also almost never use drag-n-drop. The only reason I encountered it was because I tested my clipboard improvements in x11 - just to see that other x11 events don't lock up anything :D So I don't feel very motivated to implement your suggestion. Well maybe later if I don't feel like doing anything else

valignatev: Maybe we can get it in just so it doesn't segfaults and leave a todo, idk

focus-editor:
@ileonte please feel free to change it if you want to get rid of the leak. I'm ok with having it, I feel like other things deserve more of our attention, but up to you

modules/Input/module.jai
ileonte: Sounds good to me. We can just do what was suggested above and just leak the paths for now like on Windows.

PR #162, opened 2023-09-03
Tabs render correctly when vertical scrolling

Sanian-Creations:
# Before
focus_7SZFtZzbHl

# After
focus_debug_7ICwrnnLNO


focus-editor:
Thanks. I thought I'd fixed that but apparently not

PR #163, opened 2023-09-04
move_lines_up/down now work better at EOF

Sanian-Creations:
## Summary
This bug only occured when using multicursor due to a typo in `get_whole_line_ranges_for_cursors` where it mentioned `is_last` twice, when the second one should have been `next_line_is_last`.

Code in `move_lines_up` and `move_lines_down` was reordered slightly to:
1. continue the loop earlier, 
2. never temp-copy the string more than once, and 
3. use correct length when offsetting the cursor pos after moving the lines (which was off-by-one for `move_lines_down` - I added a comment to clarify this).

# Before / After
focus_debug_ejpS7NMo7j


focus-editor:
Good catch!

PR #165, opened 2023-09-05
Add regex support in search

ileonte:
Based on the regex module `uniform` by @rluba (https://github.com/rluba/uniform). Might be slower than using something like `pcre` but it's the best thing we can get if we want to stay 100% in Jai-land (as far as I know).

Works for both project-wide as well as local file search.

There's a bit of duplicated search-related code that could probably be "collapsed" but we can do that later (the duplication existed prior to this PR as well).


rluba:
We could always start porting RE2’s conditional fast-path engines (DFA, single-pass NFA, …) into `uniform` if we want faster regex in Jai. I only ported the slower general NFA back when I started `uniform`.

ileonte:
> We could always start porting RE2’s conditional fast-path engines (DFA, single-pass NFA, …) into `uniform` if we want faster regex in Jai. I only ported the slower general NFA back when I started `uniform`.

For sure, might be a fun stream project :)

src/widgets/finder.jai
```diff
@@ -221,9 +229,28 @@ start_new_search :: (force := false) {
...
+
+        flags := BASE_REGEX_FLAGS;
+        if !case_sensitive flags |= .FoldCase;
+
+        regex, regex_is_valid = re.compile(to_string(input.text), flags);
+        if !regex_is_valid return;  // only search if the regex is valid
+
+        defer re.uninit(*regex);
+
+        matched, captures := re.match("dummy string", regex);
```
focus-editor, line 243: Dummy string?

ileonte: Yes, the string does not matter here. The idea behind that check is to "filter out" certain regular expressions that generate "null" matches. For example, `\s*` will generate such an empty capture (ie. a string with length `0`). We need to look out for those because if we do hit one it will make the searching loop infinite. For the finder widget I chose to put this check before we even initiate the search tasks (no use waking up threads just to reject the expression), for regular in-file search it happens "inline" since no tasks/worker threads are involved.

ileonte:
@focus-editor and I have discussed this PR on Discord and have agreed that I can merge it myself. Leaving this comment here just for context for when our future selves will look at this being merged by the same guy who opened the PR and wonder wtf was going on.

PR #170, opened 2023-09-06
Fix stuck keys when losing focus

valignatev:


ileonte:
Are we sure this isn't going to break things on Windows/MacOS?

Also - the Windows code path does this when the application/window gets focus back: https://github.com/focus-editor/focus/blob/main/modules/Input/windows.jai#L187 - do we want this as well?

focus-editor:
Seems to be working fine. Please report if you find bugs

PR #171, opened 2023-09-07
Improved viewport scrolling smoothness (particularly on windows)

Sanian-Creations:
- After converting all scroll-related animations to be snappy, there weren't any that were NOT snappy. So the setting was removed, animations are now always snappy. _Not delayed by 1 frame_ is a better way to describe it, all animations now begin moving the frame they are started, rather than one frame later.
- Shortened some calls to `start_animation` to be be more concise
- Came across an old procedure that was no longer used anywhere, removed that.

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -518,8 +518,8 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
 
             cursor_pos_from_left := cursor_coords[main_cursor].pos.col  * char_x_advance + text_offset.x;
             cursor_pos_from_top  := cursor_coords[main_cursor].pos.line * line_height;
-            left_target          := max(0, cast(s32)(cursor_pos_from_left - editor_width + char_x_advance * 3));
-            top_target           := max(0, cast(s32)(cursor_pos_from_top  - rect.h / 2));
+            left_target          := cast(s32) max(cursor_pos_from_left - editor_width + char_x_advance * 3, 0);
+            top_target           := cast(s32) max(cursor_pos_from_top  - rect.h / 2,  0);
             if left_target != scroll_x.target then start_animation(*viewport.scroll_x, viewport.left, left_target);
             if top_target  != scroll_y.target then start_animation(*viewport.scroll_y, viewport.top,  top_target);
 
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -480,7 +480,7 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
             using editor.viewport;
             new_target := clamp(scroll_y.target - mouse.scroll_y_delta, 0, max_y_scroll);
             if mouse.smooth_scroll {
-                start_animation(*scroll_y, top, new_target);
+                start_animation(*scroll_y, top, new_target, snappy=true);
             } else {
                 scroll_y.target = new_target;
                 top = new_target;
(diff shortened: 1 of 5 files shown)


src/draw.jai
```diff
@@ -516,15 +516,12 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
...
+            cursor_pos_from_top  := cursor_coords[main_cursor].pos.line * line_height;
+            left_target          := max(0, cast(s32)(cursor_pos_from_left - editor_width + char_x_advance * 3));
+            top_target           := max(0, cast(s32)(cursor_pos_from_top  - rect.h / 2));
             if left_target != scroll_x.target then start_animation(*viewport.scroll_x, viewport.left, left_target);
-
-            cursor_pos_from_top := cursor_coords[main_cursor].pos.line * line_height;
-            top_target := cast(s32)(cursor_pos_from_top - rect.h / 2);
-            if top_target < 0 then top_target = 0;
-            if top_target != scroll_y.target then start_animation(*viewport.scroll_y, viewport.top, top_target);
+            if top_target  != scroll_y.target then start_animation(*viewport.scroll_y, viewport.top,  top_target);
```
Sanian-Creations, line 524: Just re-ordered this for cleanliness, no functional changes.

src/draw.jai
```diff
@@ -516,15 +516,12 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
...
+            cursor_pos_from_top  := cursor_coords[main_cursor].pos.line * line_height;
+            left_target          := cast(s32) max(cursor_pos_from_left - editor_width + char_x_advance * 3, 0);
+            top_target           := cast(s32) max(cursor_pos_from_top  - rect.h / 2,  0);
             if left_target != scroll_x.target then start_animation(*viewport.scroll_x, viewport.left, left_target);
-
-            cursor_pos_from_top := cursor_coords[main_cursor].pos.line * line_height;
-            top_target := cast(s32)(cursor_pos_from_top - rect.h / 2);
-            if top_target < 0 then top_target = 0;
-            if top_target != scroll_y.target then start_animation(*viewport.scroll_y, viewport.top, top_target);
+            if top_target  != scroll_y.target then start_animation(*viewport.scroll_y, viewport.top,  top_target);
```
Sanian-Creations, line 524: Re-ordered this for cleanliness. No functional changes.

src/utils.jai
```diff
@@ -339,28 +339,14 @@ get_tmp_tabs :: (num: int) -> string {
...
-            return;
-        }
-    }
-    #if snappy {
-        // Pretend the animation started last frame so we start moving right away
-        anim.started_at = frame_time - MAX_FRAME_DT;
-    } else {
-        // Things will start moving next frame - but usually in a more smooth way
-        anim.started_at = frame_time;
-    }
```
Sanian-Creations, line 363: The only reason the non-snappy option would previously be "more smooth" was because if you had high framerate, removing MAX_FRAME_DT to "Pretend the animation started last frame" was an overestimation, so you actually went back _more_ than one frame. Now that `frame_dt` is used, which is the _actual_ time the previous frame took, the snappy version is more smooth because it no longer delays the start of the animation by one frame. This difference is less noticeable if you already have high framerate.

Sanian-Creations, line 363: The only reason the non-snappy option would previously be "more smooth" was because if you had high framerate, removing MAX_FRAME_DT to "Pretend the animation started last frame" was an overestimation, so you actually went back _more_ than one frame. Changing it to `frame_dt`, which is the _actual_ time the previous frame took, the snappy version is more smooth because it no longer delays the start of the animation by one frame. This difference is less noticeable if you already have high framerate.
For this reason the snappy option was removed completely, as now the only difference between having it and not having it was delaying the animation one frame, which is not wanted behaviour.

PR #174, opened 2023-09-14
"auto_surround_with_brackets_and_quotes" config option

hfr4:
This PR was reverted because of a compiler bug, look like it's working now.

https://github.com/focus-editor/focus/pull/154


focus-editor:
💪

Issue #14, opened 2023-05-03
Launching from command line to just quickly check out a file scans whole directory structure.


mcourteaux:
When I try to do: `./focus /home/martijn/demo.cpp`, it spends about a minute scanning my home folder, installing directory watchers.

I don't know the philosophy regarding projects and where this editor fits in besides projects, but this seems not ideal.

focus-editor:
I will investigate why it takes a minute because it shouldn't scan the folder in the sense of adding it to the workspace, and I wouldn't expect the file watcher to take a minute to add a directory (but maybe it could be the way it's done on Linux).

Currently this is how the watching of individual files is implemented - the file watcher module in Jai doesn't seem to provide the ability to watch files, only folders, this is why the parent folder is added to the file watcher.

mcourteaux:
It traverses all folder and subfolders to full depth, if that helps. It also reports a bunch of folders it can't watch, that's why I realized. Here is a snippet of the output when I try to open a file in my home folder:

![image](https://user-images.githubusercontent.com/845012/235993056-31fe553f-4bf8-4d09-bed0-e1d331e2f300.png)

mcourteaux:
Regarding what I said "I don't know the philosophy". Maybe it only makes sense to do the file watching if you have an actual project config setup?

focus-editor:
This would be inconsistent - why would the editor not react to changes made to individual files outside of the editor when there's no project setup? Also, the editor doesn't really care if you have a project or not. You could just add your project dirs to the global config if you want and it would behave exactly the same.

The correct solution to this would be to implement a way to watch individual files in a more efficient way. I will definitely explore the options when I get to this issue

mcourteaux:
Aha. I see! Makes sense. Thanks for clarifying. I like spending some time discussing issues in open source projects! I always appreciate it when my ideas are taken seriously. :smile: Keep up the nice work!

focus-editor:
This might be fixed in the next release (`0.2.0`)

When it's out please test again.

PR #176, opened 2023-09-16
Linux: Non-recursively watch parent directories of files outside of project directories

ileonte:
Not sure how correct this is, but I think it should work.
Should fix https://github.com/focus-editor/focus/issues/14

### DIFF
--- a/modules/File_Watcher/module.jai
+++ b/modules/File_Watcher/module.jai
@@ -1,0 +1,678 @@
+// At startup, call init() to create one or more watchers.
+// Start watching directories by calling add_directories(). This will watch this directory and (and all subdirectories).
+// Call deinit() to stop watching.
+
+// Periodically call process_changes to get notified of any changes that have happened since the last call.
+// Your callback will be called for every change.
+
+File_Change :: struct {
+    full_path:  string;   // The fully-pathed filename.
+	events: File_Change_Event;
+    // @ToDo: We should probably replace this with Apollo_Time
+    time_of_last_change := 0.0;  // What time it last changed. It's a bad idea to use floating-point numbers for time, but, we did that here. This time is computed by Basic.get_time(). Maybe we should let you override this.
+}
+
+File_Change_Event :: enum_flags {
+    ADDED;
+    MODIFIED;
+
+    // You might get two MOVED events if a file gets renamed, one for the old path and one for the new path.
+    // File_Watcher will add the modifiers MOVED_FROM or MOVED_TO to a MOVED event if it knows that this event is for the old path or new path, respectively.
+    // But for example on macOS you might get just two MOVED events and have to check the full_path and see if that file still exists to know which part of the move the event belongs to.
+    // Also, you might only get events for part of the move if the file gets moved from an unwatched folder into a watched folder or vice versa.
+    MOVED;
+    MOVED_FROM; // Will be combined with MOVED if we know that the MOVED event is for the old path.
+    MOVED_TO;   // Will be combined with MOVED if we know that the MOVED event is for the new path.
+
+    ATTRIBUTES_CHANGED;
+    REMOVED;
+
+    SCAN_CHILDREN; // Indicates that you might need to check all children of this directory for changes. If a directory gets added/moved we don't automatically get change notifications for all of its children.
+
+    ALL_EVENTS :: ADDED | MODIFIED | MOVED | ATTRIBUTES_CHANGED | REMOVED;
+}
+
+DEFAULT_BUFFER_SIZE :: 8000;
+
+File_Watcher :: struct (User_Data: Type = void) {
+    File_Watcher_Callback :: #type (watcher: *File_Watcher(User_Data), change: *File_Change, user_data: *User_Data);
+
+    callback: File_Watcher_Callback;
+    user_data: *User_Data;
+    events_to_watch: File_Change_Event;
+	merge_window_seconds: float;
+    watch_recursively: bool;
+	verbose: bool;
+    buffer_size: s32;
+
+    // @Stability: pool needs to be dynamically allocated since we don't control the lifetime of File_Watcher.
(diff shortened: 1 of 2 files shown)


focus-editor:
I can't test whether it works. Looks good to me, please merge if you've tested it and are happy with the change. The only thing is - not sure why we need an extra function. Wouldn't it make more sense to just pass `recursive` to the original `add_directories` call?

ileonte:
> I can't test whether it works. Looks good to me, please merge if you've tested it and are happy with the change.

I've only tested on Linux, but I believe the other platform code paths should be the same. It would be good if you could confirm it at least doesn't break the Windows build (which again, it shouldn't, but just to be safe).

> The only thing is - not sure why we need an extra function. Wouldn't it make more sense to just pass `recursive` to the original `add_directories` call?

I initially had a `recursive` parameter added to `add_directories` but I decided to add a new function because of the following: on Windows and MacOS it doesn't look like you can do non-recursive directory watches; on Linux `INOTIFY` **only** does non-recursive, so the `File_Watcher` module tries to emulate the behavior of the other platforms by walking the base directory and adding each individual sub-directory to the watcher; because the `recursive` parameter would basically be a lie on every platform except Linux I decided that a new function may be better. I have no problem changing it back to the parameter version if you think that's better.

focus-editor:
I will test on Windows tomorrow. 

It's still a bit confusing to me why `add_directories` and `add_single_directory` should have different interfaces even with the explanation you provided - it looks like the same function, just with a different number of directories.
What would be the problem if we added it to `add_directories`? I think it would be less confusing in the future.

ileonte:
I guess the logic is that `add_directories` is a 'known' interface which implies recursive watches always. Adding a parameter that 66% of the time does nothing seems wrong. With a new function we "know" from the start that it lies to us most of the time.

focus-editor:
That applies to the new function too

focus-editor:
OK I'll leave the decision to you!

ileonte:
Yes, that's what I said:
> With a new function we "know" from the start that it lies to us most of the time.

focus-editor:
We don't know that because there's no comment explaining the decision

ileonte:
Well, we "know" because we're talking about it here :))
Anyway, once you have the chance to test on Windows I'll pull this in, either with a comment explaining the new function or with the "recursive" parameter added, I'll sleep on it.

focus-editor:
Just checked it on Windows - looks like it breaks the auto-reload when an individual file is open via the command line at startup.

Still works when the file is open using Ctrl-O

ileonte:
> Just checked it on Windows - looks like it breaks the auto-reload when an individual file is open via the command line at startup.
> 
> Still works when the file is open using Ctrl-O

I just booted in Windows and tested, it seems to work fine for me. The way I tested is as follows:
* have the following paths in `[workspace dirs]` in the global config: `D:/dev/focus`, `D:/dev/langs/jai/modules`
* create `e:/test/brrap.txt` (outside of the workspace dirs)
* run `focus e:/test/brrap.txt`
* open `e:/test/brrap.txt` in notepad, make changes and save - Focus auto-reloads the file just fine
* in Focus open a workspace file from the Ctrl-P menu (for example `src/main.jai`). Open the same file in Notepad, make changes and save - Focus also auto-reloads this fine.

focus-editor:
Apparently the bug only happens when the file you open is in the disk root - here's a screencast. If you open a dir not in the root it works
reload

ileonte:
I'm not sure what to tell you, this works fine for me. Here is how that looks on my machine: https://www.youtube.com/watch?v=i5uTW0XVp9M
Note that for me both my branch and the official `0.2.0` binary from Github sometimes need to have some event trigger (like a mouse move/hover) to trigger the refresh, you can see it happening for both in my video. Other than that, they both work the same for me.

focus-editor:
Maybe it's different on Windows 11. I'll check it one more time tomorrow. If it still fails, we'll merge it anyway since the benefits outweigh the bug

ileonte:
Regardless of Windows differences, both `add_directories()` (https://github.com/ileonte/focus/blob/linux-support/modules/File_Watcher/module.jai#L111) and `add_single_directory()` (https://github.com/ileonte/focus/blob/linux-support/modules/File_Watcher/module.jai#L127) do the same thing on Windows, I really see no reason for there to be any difference.

Is `W:` on your machine a network mount or a local drive? Do you see the same issue with `C:`? (in my example `E:` is a local disk).

focus-editor:
In your video you reopen the file using the open file dialog - I just did the same and can confirm it starts working after that, but not before. Try it fresh without reopening - I'm not lying, it doesn't refresh in your branch but does refresh in main.
Tested multiple times.

I think it's best to fix that now and not later when we forget the details - because eventually people are going to hit this bug and will be angry and will demand immediate fixes :)

ileonte:
Again, not sure what to tell you, that works fine for me still: https://www.youtube.com/watch?v=_tB6s9vxddM

The only "issue" I see with the autoreload on my system is that sometimes I have to hover the window to trigger it, but that happens with my local build as well as the `0.2.0` binary from GitHub. This also happens on Linux, I suppose it's a byproduct of the way we wrote the event loop (`File_Watcher` does nothing that would wake `platform_wait_message()`).

As I said in my previous message, the code path for Windows has not changed in this PR as both `add_directories()` and `add_single_directory()` do the same thing in this case:

- https://github.com/ileonte/focus/blob/linux-support/modules/File_Watcher/module.jai#L112
- https://github.com/ileonte/focus/blob/linux-support/modules/File_Watcher/module.jai#L127

The main change this PR introduces is replacing a couple `add_directories()` calls with `add_single_directory()` (https://github.com/focus-editor/focus/pull/176/commits/56db907af979cd267be8e8ef35af8128497a88b4#diff-e78e8beb03cc61529ead3f4fb4746a34f76bfedfa9e099ce4f83e14b1a0a9074), but as far as I can tell these are equivalent on Windows (as shown above).

At this point I'm not sure what I'm supposed to fix (or even how to debug this).

focus-editor:
It's going to remain a mystery then

PR #178, opened 2023-09-19
Create cursors with mouse middle button

hfr4:
Drag middle mouse button to create multiple cursors (+ config option)

https://github.com/focus-editor/focus/assets/30574115/acb334f3-eb9b-4ebf-bd02-ff507b571fda

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -447,7 +447,7 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
         // Activate editor
         make_editor_active(editor_id);
         if mouse.left.just_pressed then redraw_requested = true;  // so that we don't draw 2 active editors
-    } else if ui.active == ui_id && (mouse.middle.just_pressed || mouse.middle.is_dragging) && config.settings.create_cursors_with_middle_mouse_button {
+    } else if ui.active == ui_id && (mouse.middle.just_pressed || mouse.middle.is_dragging) {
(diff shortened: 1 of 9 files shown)


focus-editor:
Looks awesome! 

Sorry, I'm busy and won't be able to review it properly for a while. One thing - I think we don't need a config option in this case. The middle mouse button is not easy to press accidentally, so I think it's OK to just have it enabled always.

Later we're going to have an option which can disable all multi-cursor functionality at once for those who don't need it. All multi-cursor users can benefit from this I think.

focus-editor:
I was going to say "fair enough" for your comment on linux users but someone distracted me and I forgot :)

Let's see if someone complains. I'm going to merge it in the mean time. Thanks!

src/draw.jai
```diff
@@ -444,8 +447,85 @@ draw_editor :: (editor_id: s64, main_area: Rect, footer_height: float, ui_id: Ui
...
+        editor.refresh_highlights = true;
+
+        // Activate editor
+        make_editor_active(editor_id);
+        if mouse.middle.just_pressed then redraw_requested = true; // so that we don't draw 2 active editors
     }
-    if !mouse.left.is_down || mouse.left.just_released {
+
+    if (!mouse.left.is_down || mouse.left.just_released) || (!mouse.middle.is_down || mouse.middle.just_released) {
         selection_mode.type = .chars;
```
focus-editor, line 529: This broke different selection modes. Please fix @hfr4

PR #179, opened 2023-09-19
Add non-recursive `File_Watcher` for files outside of project directories

ileonte:
This is a re-implementation of https://github.com/focus-editor/focus/pull/176 

Instead of modifying the `File_Watcher` module we now use two watchers:
- a recursive watcher for project directories
- a non-recursive watcher for the parent directories of files outside of project directories


src/workspace.jai
```diff
@@ -445,29 +451,51 @@ file_change_callback :: (watcher: *File_Watcher(void), change: *File_Change, use
...
+        assert(watcher == *watchers.project_watcher);
+
+        // Most commonly we will receive only a single MODIFIED event for a single file (after saving a buffer)
+        // In this case we don't want to rescan anything, but only if no other events were received in the same frame
+        if change.events == .MODIFIED && !watchers.project_file_modified {
+            watchers.project_file_modified = copy_temporary_string(change.full_path);
         } else {
-            dir = get_parent_dir_path(change.full_path);
+            // There are other events - maybe rollback and add dir to the rescan queue
+            if watchers.project_file_modified {
```
focus-editor, line 480: A bug right here. The `> 0` was intentional. Hard to catch by just reviewing

ileonte: I'm not sure I understand what you mean. Both of these branches are taken:
```    s := "test";
    if s           print(">>> YEP!\n");
    if s.count > 0 print(">>> ALSO YEP!\n");
```

focus-editor: Look at line 483. `count` is set to `-1`, in which case if you compare it with `if s`, the branch will be taken but it's not what we want

ileonte: Right, I remember seeing that and the forgetting about it almost instantly. There has to be a less 'sneaky' way to track that state, I don't think anyone would expect a string with a length of `-1`.

PR #182, opened 2023-09-21
Swap selections

hfr4:
Swap two selections


focus-editor:
I'm going to merge it, however I doubt it's useful often enough to get a default shortcut - I'll remove it so that Alt-S is free for something else in the future. Also note that any shortcut of the form `Alt-<key>` is going to be incorrect on macOS because that's how special characters are typed there

PR #185, opened 2023-09-22
remove State struct from cursor

hfr4:


focus-editor:
Oh yeah there's no need for this struct any more

PR #187, opened 2023-09-24
Custom UI font

hfr4:
Configure ui font sizes and families from config file.


focus-editor:
Sorry I have a big queue of stuff I need to do before I can look at this, I will have to put another pause on PR reviews.

focus-editor:
This is what I see when I run the code in this branch, even with no directories in the workspace:

image

PR #198, opened 2023-09-29
Simple Word Completion

polgartom:
Maybe too complicated in some places, because I don't know the source code fully, so maybe you can do it a simpler way. I've been using this for the past 2 months, and it's not as annoying as LSP is (at least for me). Maybe we need to do some cleanup on this, but initially, I just want to create something that is good and simple.

**The extra features:**

- Checks the distance between your FIRST cursor and the found word (or subwords) and provides results accordingly.
- If your starting word slice begins with an uppercase letter, it will prioritize words starting with uppercase letters, followed by words starting with lowercase letters.
- You can search for subwords (this feature only works with snake_case words). Subword search combines the two features mentioned above and it first lists words without underscore characters, then words with underscore characters.


focus-editor:
I will review this next week when I'm back from my trip and have cleared the backlog of stuff I have - will do before the next release.

ustance:
Checked it out a few days ago, was working pretty nice. Hope it gets merged, it's a solid step towards Jails/lsp support :)

focus-editor:
Sorry for taking this long. I haven't forgotten about it - been sick and didn't work on the editor at all.

focus-editor:
This is getting delayed since I need to take a very careful look at it and we barely have the time to finish the remaining things for the release

polgartom:
This is a simple stuff which search for words in the buffer. I cannot live without that but probably you would refact the whole thing as it is with more optimization.

focus-editor:
OK finally I'm ready to look at this. Thanks for waiting

focus-editor:
I've tried it once again. The reason I needed to take a careful look at it was because when I tried it at first I wasn't very happy with the behaviour and felt the need to redo it.

I hope you're OK if I use your code as a reference and implement the behaviour I'd intended to implement originally.

polgartom:
If you able to use anything from it then just go ahead. :)

focus-editor:
I haven't forgotten about it, just super busy. Completions should be coming in 0.4.0 (but we haven't shipped 0.3.6 yet :( )

polgartom:
Hi! Oh thats gonna be cool it has so many good features already. Yeah, well I just try to keep it uptodate.

PR #199, opened 2023-09-29
Add single empty line comment (just like in vscode)

polgartom:
You can add single empty line comment.
If you select multiple lines, whether they're empty or not, at once, it won't work.
As I said.. just like in vscode.


polgartom:
I did more testing and very strange thing is happening in release build with it.

PR #206, opened 2023-10-13
Fix open file search dialog not allowing "/" keystrokes

### LINKED ISSUE #203: Typing `/` doesn't do anything in file search
**Describe the bug**
When hit ctrl-p and typing `/` nothing happens

**To Reproduce**
1. Hit ctrl-p to open file search (default config)
2. Type `/`

**Expected behavior**
When `/` typed, it appears in the search box and the string is matched against path

**Additional context**
Arch linux on x11.

SentientCoffee:
Fixes #203.

This wasn't an issue on Windows because of the way the Input module works I think...It seems that the "/" key press generates `TEXT_INPUT` events on Linux but not on Windows. Could be worth mailing in? Would need to do more research to be sure though.


SentientCoffee:
Found the root cause:
20231013_044607_Screenshot

`event.text_input_count` is never set on Windows, which explains the discrepancy I mentioned above (it does generate the `TEXT_INPUT` event, but we depend on the `text_input_count` to know how many text inputs to ignore).

PR #210, opened 2023-10-21
XML support

arjenwitteveen:
As promised on Discord, here is a pull request that adds highlighting and basic indentation support for XML (and HTML). It's nobody's favorite language, but the preponderance of files in this format makes it useful to support.

The code is by no means an implementation of the full XML or HTML specification, but aims to tokenize files in a way that is performant and gives good results for most use cases (including things like XML Schema and Document Type Definition). The tokenizer is fairly lenient, so it won't mark many things as errors.

For performance testing (apart from profiling) my benchmark was a 13 megabyte HTML file; editing this file works decently in release mode with the current implementation. There's some character lookup tables generated at compile-time that speed up some of the hot code paths.

Some further notes:

- Using `toggle_block_comment` works, but not `toggle_comment`; the latter does not support comments that are delimited by both a comment start and end token, like XML has.

- Indentation works by treating opening tags as an 'open brace' and closing tags as a 'closing brace'. This works fine (including with self-closing tags) as long each tag is on its own line (or opening and closing tag are on the same line); if the layout of tags in the file is messy, the indentation will be messy too.


focus-editor:
Thanks for this. Would you be able to fix the conflicts caused by another PR I just merged, then I'll merge it?

arjenwitteveen:
Just merged the latest changes. Didn't see any conflicts though!

PR #219, opened 2023-11-19
fix crash when writing outside config section

jmpstar:
This fixes a crash when writing something between version number and [[workspace]] section of the config

The editor crashes with array bounds check error when writing content between the Version number and [[workspace]] section of the config.
This happens both for the global configuration as well as the project configuration.

Let me know if this should be a hard fail. For now it highlights the error and logs a warning.
https://github.com/focus-editor/focus/assets/6022262/e4db552a-796f-4e97-b4fe-86cb4322aa03

### DIFF
--- a/src/config_parser.jai
+++ b/src/config_parser.jai
@@ -87,7 +87,7 @@ parse_config :: (name: string, filename: string, file_data: string, for_highligh
         // Check if we have an active_section
         if active_section == .none {
             add_highlight(*parser, 0, line.count, .error);
-            error_msg := log_parser_error(handler, "Trying to add config outside of config section: '%'. This is an error.", line);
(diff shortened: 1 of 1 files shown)


src/config_parser.jai
```diff
@@ -84,6 +84,14 @@ parse_config :: (name: string, filename: string, file_data: string, for_highligh
             if handled continue;
         }
 
+        // Check if we have an active_section
+        if active_section == .none {
+            add_highlight(*parser, 0, line.count, .error);
+            error_msg := log_parser_error(handler, "Trying to add config outside of config section: '%'. This is an error.", line);
```
focus-editor, line 90: Maybe rephrase to `"Line outside of any section: '%'. This is an error."` to be more consistent with the error message below?

Otherwise all good, thanks for the fix

jmpstar: sure, done

PR #220, opened 2023-11-20
Add basic yang syntax support

jmpstar:
As hinted at in the discord, this PR adds basic syntax highlighting support for the YANG Modeling lanuage (https://datatracker.ietf.org/doc/html/rfc7950) as well as support for the commands `toggle_comment` and `toggle_block_comment` for the language.

This version of the parser is not context aware of the statements so does not support highlighting of custom types defined in the model but does a good job of syntax highlighting. Statement awareness would need to be added for more advanced language support like linting.

This is an example of how it looks: image
By default `TYPE_KEYWORDS` are highlighted with default code color to avoid angry fruit salad as jon would put it.

### DIFF
--- a/src/langs/yang.jai
+++ b/src/langs/yang.jai
@@ -1,0 +1,283 @@
+highlight_yang_syntax :: (using buffer: *Buffer) {
+    tokenizer := get_tokenizer(buffer);
+
+    while true {
+        token := get_next_token(*tokenizer);
+        if token.type == .eof break;
+
+        using tokenizer;
+
+        color := COLOR_MAP[token.type];
+        memset(colors.data + token.start, xx color, token.len);
+    }
+}
+
+#scope_file
+
+get_tokenizer :: (using buffer: Buffer) -> Tokenizer {
+    tokenizer : Tokenizer;
+
+    tokenizer.buf = cast(string) bytes;
+    tokenizer.max_t = bytes.data + bytes.count;
+    tokenizer.t     = bytes.data;
+
+    return tokenizer;
+}
+
+get_next_token :: (using tokenizer: *Tokenizer) -> Token {
+    t = eat_white_space(t, max_t);
+
+    token : Token;
+    token.start = cast(s32) (t - buf.data);
+    token.type  = .eof;
+    if t >= max_t return token;
+
+    start_t = t;
+
+    char := << t;
+
+    if is_alpha(char) {
+        parse_identifier(tokenizer, *token);
+    } else if is_digit(char) {
+        parse_number(tokenizer, *token);
+    } else if char == {
+        case #char "\"";        parse_string_literal(tokenizer, *token, char);
+        case #char "'";         parse_string_literal(tokenizer,*token, char);
+        case #char "/";         parse_slash_or_comment(tokenizer, *token);
+
+        case #char ";";         token.type = .punctuation; token.punctuation = .semicolon; t += 1;
+        case #char "{";         token.type = .punctuation; token.punctuation = .l_brace;   t += 1;
+        case #char "}";         token.type = .punctuation; token.punctuation = .r_brace;   t += 1;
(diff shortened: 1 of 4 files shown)


src/langs/yang.jai
```diff
@@ -0,0 +1,283 @@
...
+
+get_tokenizer :: (using buffer: Buffer) -> Tokenizer {
+    tokenizer : Tokenizer;
+
+    tokenizer.buf = cast(string) bytes;
+    tokenizer.max_t = bytes.data + bytes.count;
+    tokenizer.t     = bytes.data;
+
+    return tokenizer;
+}
```
focus-editor, line 25: This function is now available in the `common.jai` file, so no need to redefine it here unless you need a special `Tokenizer` (which in this case doesn't seem to be required)

src/langs/yang.jai
```diff
@@ -0,0 +1,283 @@
...
+    return token;
+}
+
+parse_identifier :: (using tokenizer: *Tokenizer, token: *Token) {
+    token.type = .identifier;
+
+    identifier_str := read_identifier_string(tokenizer);
+    
+    // need to do this if we want to use the same enum logic as in other parsers, not pretty but get's the job done.
+    sanitized_string := replace(identifier_str, "-", "_");
```
focus-editor, line 70: This line leaks the string because `replace` allocates on the heap. I guess here you could just do a `push_allocator(temp);` at the top of the function and then it'll use temporary storage instead.

src/langs/yang.jai
```diff
@@ -0,0 +1,283 @@
...
+
+
+Tokenizer :: struct {
+    buf: string;
+    max_t:   *u8;
+    start_t: *u8;  // cursor when starting parsing new token
+    t:       *u8;  // cursor
+
+    last_tokens: [2] Token;  // to retroactively highlight functions
+}
```
focus-editor, line 165: No need to redefine `Tokenizer` either, since the default one from `common.jai` would fit just as well. Just remove it and I think it'll just work (as well as the `get_tokenizer` function

src/langs/yang.jai
```diff
@@ -0,0 +1,283 @@
...
+    while t < max_t {
+        c := <<t;
+        if is_alnum(c) || c == #char "-"  { t += 1; continue; }
+        break;
+    }
+    if t >= max_t then t = max_t;
+    ret.count = xx (t - ret.data);
+
+    return ret;
+}
```
focus-editor, line 155: This is another common function which can just be deleted.

src/langs/yang.jai
```diff
@@ -0,0 +1,283 @@
...
+    return xx result;
+}
+
+eat_white_space :: inline (start: *u8, max_t: *u8) -> *u8 {
+    t := start;
+    while t < max_t && is_white_space(<<t) {
+        t += 1;
+    }
+    return t;
+}
```
focus-editor, line 280: Guess what? :)

P.S. `eat_white_space` has changed its format in `common.jai`, it now accepts only `Tokenizer` as an argument and returns nothing. Please check how it's done in other places.

You seem to have based your implementation off of an old one which has since been refactored.

focus-editor (changes requested):

src/langs/yang.jai
jmpstar: For this one there is a slight difference to the version in common.jai as we also need to match `-` as part of identifiers in Yang. Depending on how you think about supporting other languages, does it make sense to integrate a facility in the `common.jai` version of `read_identifier_string` to steer this?

One option I could see is maybe introduce a parameter in context similar to what happens for selection commands and word matching with `context.underscore_is_part_of_word`.

Another one could be to pass a list of additional valid chars that should be checked?

Or we leave it as is and redefine it here? What do you think?

focus-editor: If it's different from common then yes it should be defined here

PR #234, opened 2023-12-13
Add Lua 5.4 syntax highlighting and indentation tokenization

SentientCoffee:


src/langs/lua.jai
```diff
@@ -0,0 +1,590 @@
...
+    identifier.data = t;
+
+    while t < max_t && is_alnum(t.*) {
+        t += 1;
+    }
+    if t >= max_t then t = max_t;
+    identifier.count = t - identifier.data;
+
+    return identifier;
+}
```
focus-editor, line 414: Can the generic function be used here instead? If the only reason for having it is the custom tokenizer, we could make it inherit from `Tokenizer` with `#using #as` and then use the generic function (here and in other places)

src/langs/lua.jai
```diff
@@ -0,0 +1,590 @@
...
+
+eat_white_space :: (using tokenizer: *Lua_Tokenizer) {
+    while t < max_t && is_white_space(t.*) {
+        t += 1;
+    }
+}
+
+is_hex :: inline (c: u8) -> bool {
+    return is_digit(c) || (c >= #char "a" && c <= #char "f") || (c >= #char "A" && c <= #char "F");
+}
```
focus-editor, line 430: We already have a function for that don't we?

src/langs/lua.jai
```diff
@@ -0,0 +1,590 @@
...
+
+is_white_space :: inline (char : u8) -> bool #no_aoc {
```
focus-editor, line 432: Same question here. If this one is different for some reason, then please add a comment explaining why we had to copy it.

focus-editor (changes requested):

SentientCoffee:
The problem with the review was just that I copy pasted the file to keep it all contained in one file, I can delete the unneeded procs and resubmit.

src/langs/lua.jai
SentientCoffee: See below comment.

src/langs/lua.jai
SentientCoffee: See below comment.

src/langs/lua.jai
SentientCoffee: I've changed the `Lua_Tokenizer` to have `using #as base: Tokenizer`.

focus-editor:
Thanks @SentientCoffee !

PR #236, opened 2023-12-14
fast first index

mathaou:


focus-editor:
I tested this change and it crashes for me as soon as I try to select a word using Ctrl-D (which is when this function is called)

matthewfarstad:
I had it being called at select all index as well with CTRL+SHIFT+A and it was working then. Didn't know CTRL+D also called so I'll try that as well and fix

focus-editor:
Sorry man, I've tried it again and it crashed again. Honestly I don't think using SIMD here is really worth it at this point, because this function is used rarely and SIMD is not without a cost.

I'm closing this PR because I've spent enough time on this and don't want to spend any more. I would appreciate it if in your next PRs you tested it thoroughly yourself before submitting. Reviewing PRs is a huge time and energy drain for me, I didn't realise that until I started doing it for Focus

PR #238, opened 2023-12-16
[+] .bat file .sh equivalents for when working with MacOS

SerialDev:


focus-editor:
Actually I've been meaning to get rid of the batch scripts as well. `jai first.jai` is all you really need, and having heavy-ish scripts is a little bit against jai's philosophy.

Especially now that there's a build system in Focus, they are not needed. Please feel free to use scripts locally but we probably won't be merging them here, sorry.

PR #239, opened 2023-12-20
Persistent search highlights

amasondev:
Makes search highlighting stay visible after closing the search bar. This is similar to vim's `hlsearch` option.

Relevant discord msg: https://discord.com/channels/1106007785193345104/1126291418202251368/1153893543820140646

Question: should this be default behaviour or behind a config option?

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -288,9 +288,7 @@ active_editor_handle_event :: (editor: *Editor, buffer: *Buffer, event: Input.Ev
     if buffer.dirty {
         rescan_for_lines(buffer);
         if line_wrap_is_active(editor) then rescan_for_wrapped_lines(editor, buffer);
-        if editor.search_bar.results {
-            search_and_update_results(editor, buffer, jump = false);
-        }
+        if editor.search_bar.results then search_and_update_results(editor, buffer, jump = false);
     }
     for * editor.cursors put_cursor_in_valid_spot(it, buffer);
     if editor.cursors.count > 1 then organise_cursors(editor);  // sort and maybe merge overlapping cursors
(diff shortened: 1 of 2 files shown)


src/editors.jai
```diff
@@ -285,6 +288,9 @@ active_editor_handle_event :: (editor: *Editor, buffer: *Buffer, event: Input.Ev
     if buffer.dirty {
         rescan_for_lines(buffer);
         if line_wrap_is_active(editor) then rescan_for_wrapped_lines(editor, buffer);
+        if editor.search_bar.results {
+            search_and_update_results(editor, buffer, jump = false);
+        }
```
focus-editor, line 293: Please put on one line with a `then` since we do it in this codebase when we can

src/editors.jai
```diff
@@ -376,6 +382,9 @@ active_editor_type_char :: (char: u32) {
     buffer.last_edit_time = frame_time;
     editor.cursor_moved = .unimportant;
     cursors_start_blinking();
+    if editor.search_bar.results {
+        search_and_update_results(editor, buffer, jump = false);
+    }
```
focus-editor, line 387: Same here

focus-editor (changes requested):
We need to refresh results in at least 2 more places - in `refresh_open_buffers` if the buffer is still dirty, and in `refresh_buffer_from_disk`, otherwise when a buffer is modified elsewhere the editor won't update the highlights.

amasondev:
> We need to refresh results in at least 2 more places - in `refresh_open_buffers` if the buffer is still dirty, and in `refresh_buffer_from_disk`, otherwise when a buffer is modified elsewhere the editor won't update the highlights.

Good catch. The change to refresh_all_editors_for_buffer catches both of these.

focus-editor:
This will now be disabled by default to avoid confusion. Please enable manually using `persist_local_search_results: true`

PR #240, opened 2023-12-20
Fix generic dialog not correctly drawing the input.

amasondev:
Was drawing the input for the commands dialog.


focus-editor:
Ouch

PR #241, opened 2023-12-20
highlight matching brackets and jump to matching brackets

amasondev:
Last PR for a while I swear :) I've been using this locally for a while and happy with it. Left it off by default.

Only questionable thing I can think of here is whether to implement this for input widgets as well. Stuffed a TODO there.

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -64,6 +64,7 @@ highlight_line_with_cursor:             false
 draw_indent_guides:                     false
 auto_surround_with_brackets_and_quotes: false
 auto_close_brackets:                    false
+highlight_matching_brackets:            false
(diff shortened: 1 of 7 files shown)


focus-editor:
I'm calling a PR freeze until we do a release.

Before this feature is implemented we'd need to rename colors to tokens and use the token information to find the matching brackets, otherwise we'll match wrong things.

Also not sure I like a switch statement instead of a bunch of `||`. Will review after release but it will require other changes.

Also, please, can you conform to the surrounding style when submitting a PR.

amasondev:
> Before this feature is implemented we'd need to rename colors to tokens and use the token information to find the matching brackets, otherwise we'll match wrong things.

Makes sense, will withdraw this one then.

PR #243, opened 2023-12-25
Improve assert/crash report

Sanian-Creations:
## Changes
- `focus-crash-report.txt` will now include a stack trace, if available (see note). This works in release.
- The same crash report is also included in the session log `log_error.txt` (since `focus-crash-report.txt` gets overwritten every time a crash occurs)
- For debug mode: The stack trace is now printed to the console *before* the assert messagebox shows up, so you can view it even if the console window closes with the application.

## Note
One issue with printing the stack trace to the crash report is that it really only works for assertions, not for bounds-checks and null-dereferences etc. The handlers for those (like `__array_bounds_check_fail`) are `#no_context`, and pushing a new context creates a new stack trace.
It's possible to get the stack trace from the OS like `Runtime_Support_Crash_Handler` does, but that may be a bit too involved for now. Currently it will always print the stacktrace to the console, we just can't always save it to a file.

### DIFF
--- a/modules/Runtime_Support.jai
+++ b/modules/Runtime_Support.jai
@@ -686,6 +686,7 @@ report_error_in_yer_face :: (message: string, args: .. Any) {
     #if OS == .WINDOWS {
         W :: #import "Windows";
         W_Utf8 :: #import "Windows_Utf8";
+        W.ShowCursor(1);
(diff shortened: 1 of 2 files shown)


focus-editor:
Thanks, working very well

PR #245, opened 2023-12-25
Simplify, remove pointless #no_reset

Sanian-Creations:
`#no_reset` should be used when _modifying_ a global variable at comptime. Here it was just being assigned a value generated by `#run`, so it didn't need to be there.
Change was introduced in f620a34baff367a4ad2b72869e9b89e9e1553d60


focus-editor:
I had already changed this piece of code locally before seeing this
FYI we're in a PR freeze now, everything will be reviewed after the release.

PR #247, opened 2023-12-26
switch_to_project no longer prematurely loads config

Sanian-Creations:
# Issue repro
- Create New File
  - type some stuff in there
- Switch To Project
  - Save
    - hit Escape
- Open Project Config
  - note the config is that of the selected project, but the workspace has not been reloaded because you canceled the save.

# Fix
`open_selected_project` no longer instantly loads the new project config, but does so after the save dialog completes. To do that, `show_unsaved_buffers_dialog` and `add_buffer_to_save_queue` have gotten a new mode/action: `.switch_project` to call on once all files are saved.

### Sidenote
I'm not a fan of the way dialogs work right now. `open_selected_project` has the potential to call a chain of 3-4 different kinds of dialogs, but they feel a little too loosely connected. Preferably a change down the line makes it so that dialogs are saved in some kind of stack so that when the save dialog closes it knows to return control to the open_project_dialog that initially called forth the save dialog. Then the open_project_dialog could decide to reload or not, but such a change would affect other dialogs too, I'd rather discuss something like that before implementing it.

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -1132,6 +1132,7 @@ process_buffer_save_queue :: () {
             case .do_nothing;           // yay, do nothing
             case .exit_program;         should_quit = true;
             case .reload_workspace;     should_reload_workspace = true;
+            case .switch_project;       should_switch_project = true;
         }
     }
 }
@@ -1403,6 +1404,13 @@ Indentation_Token :: struct {
 }
 
 
+After_Save_Action :: enum {
+    do_nothing;
+    exit_program;
+    reload_workspace;
+    switch_project;
+}
+
 #scope_file
 
 // A queue to allow us to use the in-editor save dialog for multiple buffers.
@@ -1413,8 +1421,3 @@ buffer_save_queue: struct {
     after_save: After_Save_Action = .do_nothing;
 };
 
-After_Save_Action :: enum {
-    do_nothing;
-    exit_program;
-    reload_workspace;
-}
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -1132,7 +1132,7 @@ process_buffer_save_queue :: () {
             case .do_nothing;           // yay, do nothing
             case .exit_program;         should_quit = true;
             case .reload_workspace;     should_reload_workspace = true;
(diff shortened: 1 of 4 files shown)


focus-editor:
The way dialogs are done is not final, but rather than implementing a
generic system that would handle hypothetical situations I would prefer to
solve a real problem that we have.

Maybe instead of a stack we'll just have a separate editor state, in which
it will make sure to properly chain dialogs and do proper cleanup on
premature exit etc

focus-editor:
This is an issue I was aware of but the time didn't feel right to attempt a proper fix. I'm not sure I like the proposed fix either (as you probably are yourself given the current implementation)

With your permission I'll leave this PR open until after the release and then we'll use it to properly examine the problem and simplify the workspace reloading stuff as well as save buffer queues.

I think we'll try to do it by introducing editor states, like the following: default, closing, switching_to_project, saving_buffers etc. This should make the code more clear and more robust.

Sanian-Creations:
Such states would need to hold a little more info than just a single enum though, for example switching to a project also requires a string for the project to switch to. And some of the states you list are things that may need to happen consecutively, like saving buffers may precede a variety of different actions. A single enum cannot hold that information, so you'd need something at least a *little* more complex than that. 

Maybe not for the dialogs themselves, but I really think a stack or queue would work well for keeping track of the actions that need to happen.

To switch to a project you could put the following actions on a stack:
- reload the workspace.
- but first load a new config.
- but first save all modified files.

A single action could be an enum, but having a list of them to know in which order they need to happen just seems like it could help simplify things a lot.

focus-editor:
Not sure where you got the idea that a single enum was going to be used. I was just trying to describe the general idea.
I'm just going to stop talking.

Sanian-Creations:
Sorry about that, just trying to brainstorm a little

focus-editor:
I've thought about it carefully and ended up going with a simple deferred action stack like you suggested.

For the tasks we currently have introducing explicit states is not required, mainly because all transitions are linear right now and there's no cleanup required when you cancel a chain halfway. We might have to do it later but right now it's not needed.

Issue #248, opened 2023-12-27
add cursors to line starts crashes on EOF


Sanian-Creations:
### Bug description
The command "Add Cursors To Line Starts" does not work properly (it can delete the cursor that is there) on the last line of a file and can trigger an assert in some circumstances.

### Repro
1. open a new file
2. insert a single newline
3. run the command "Add Cursors To Line Starts" (or "Ends")

focus-editor:
Please check if there are still issues with this. Thanks

PR #249, opened 2023-12-27
Fix add cursors to line starts

Sanian-Creations:
Fixes Issue #248 

### Changes
Added a new editor utility proc `get_real_line_numbers_for_cursors`, similar to `get_whole_line_ranges_for_cursors`, which returns ranges of line numbers rather than ranges of character offsets. The reason here being that the other proc was already calculating line numbers in order to turn them into offsets, which `add_cursors_to_line_starts` was then trying to turn back into line numbers. This new proc performs less work and simplifies the implementation of `add_cursors_to_line_starts`.

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -2588,7 +2588,7 @@ Line_Number_Range :: struct {
 get_real_line_numbers_for_cursors :: (editor: Editor, buffer: Buffer) -> [] Line_Number_Range /* tmp */ {
     line_num_ranges: [..] Line_Number_Range;
     line_num_ranges.allocator = temp;
-    line_after_prev_range: type_of(Line_Number_Range.last_line_num);
+    line_after_prev_range: s32;
 
     for cursor : editor.cursors {
         selection := get_selection(cursor);
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -2476,16 +2476,13 @@ remove_additional_cursors :: (editor: *Editor) {
 add_cursors_to_line_starts :: (editor: *Editor, buffer: Buffer) {
     if buffer.bytes.count <= 0 return;
 
-    line_ranges := get_whole_line_ranges_for_cursors(editor, buffer);
-    array_reset_keeping_memory(*editor.cursors);  // reset after getting ranges (obviously)
+    line_num_ranges := get_real_line_numbers_for_cursors(editor, buffer);
+    array_reset_keeping_memory(*editor.cursors);  // reset *after* getting ranges
 
-    for * line_range : line_ranges {
-        line_num := offset_to_line(editor, buffer, line_range.start);
-        while true {
-            pos := buffer.line_starts[line_num];
-            if pos >= line_range.end break;
-            array_add(*editor.cursors, Cursor.{ pos = pos, sel = pos, col_wanted = -1 });
-            line_num += 1;
+    for range : line_num_ranges  {
+        for range.first_line_num..range.last_line_num {
+            offset := buffer.line_starts[it];
+            array_add(*editor.cursors, Cursor.{ pos = offset, sel = offset, col_wanted = -1 });
         }
     }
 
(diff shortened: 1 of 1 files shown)


focus-editor:
I understand that the proposed solution is more efficient than the original one, however I don't think this efficiency is worth adding a new nontrivial utility function which is only used once, in a situation when we don't really care about efficiency.

Instead of doing this I've tried to fix #248 by handling the situation when we have no cursors at the end (will push now). Please let me know if there are still issues (I didn't fully understand what you meant by it removing the cursors).

Sanian-Creations:
It's not quite fixed, if you have a file with this content:
```
a|
b|
c|
```
Imagine the `|` are cursors. There is no newline after `c`. Running add_cursors_to_line_starts will work fine.
Now remove the `c`.
```
a|
b|
|
```
Run the command again. The cursor that was on the last line gets deleted, leaving only the 2 above. This is what originally caused the assert to fail, the cursor on EOF would get deleted leaving you with 0 cursors.

Sanian-Creations:
Actually, it can even crash still. For some reason the order in which you create the cursor matters.
```
a|
b

```
Start the cursor at the position and then Alt-Shift-DownArrow to create 2 additional cursor below it to re create the example above. It will trigger an assert in draw.jai
If you create the cursor in another order and do the same it may not trigger.

focus-editor:
What you're describing in the first comment isn't an issue though. There's no line at the end when you delete the `c`, so no cursors get added there. This is the exact same behaviour that is in VSCode, for instance.

Re: the 2nd comment - I couldn't get it to crash doing what you said. If you can do it consistently, please give more information, i.e. what assert gets triggered etc.

Do note that debug builds use an unmapping allocator and memory debugger is turned on, which can cause spurious crashes sometimes for reasons outside of the focus codebase.

focus-editor:
Ok I think I managed to crash it, will fix

Sanian-Creations:
Look, if you don't want the utility function cuz it's only used in one place, I get that. But I really think that using `get_whole_line_ranges_for_cursors` to derive line numbers is doing unnecessary work and making things harder than they should. It's getting line numbers, turning them into offsets, and then after we call it we turn them back into line numbers. I could integrate the content of the proc I made into `add_cursors_to_line_starts` and that would work too.

focus-editor:
Mate. I appreciate it that you're contributing code and reporting issues and I'm grateful for that. However, I've spent a significant chunk of my morning today arguing over how something should or shouldn't be implemented with you, and that time surely would be better spent elsewhere. 

It doesn't feel right for me to just ignore people and do my thing, or just blindly accept whatever people are proposing, so I have to go and spend time reviewing stuff.

I will explain once again why I think we shouldn't do what you're suggesting here:
1. It will be more code than there currently is.
2. It will introduce more mental complexity when reading code (as opposed to using a well-established get_whole_ranges function which is used in many places and it's clear what it does). It's straightforward to understand - get the ranges, then add cursor to each line_start. Do note that we only call `offset_to_line` once per range (and usually you will only have one range).
3. Even if the existing solution is less efficient that a direct solution would be, *we don't care in this case*.

If you disagree with the technical decisions that I make, it's perfectly OK. If you really want to implement your things, it would be best if you did that in your fork. I like to design how I like to design, and arguing over implementation is not something that I think we should be spending time on. I hope you understand. I want to maintain good relationship and I don't want to upset people.

Sanian-Creations:
Well you fixed the problem, so it's fine. Not trying to waste anyone's time here, sorry if it turned out that way. 

I'll keep the points you listed above in mind when making future changes, maybe run them by people on discord first before making a PR though, to avoid stuff like this from happening again. 

Cheers!

focus-editor:
Thanks for your understanding. It's not you wasting my time, it's me wasting my time because I can't just ignore people, while it would really be more efficient to just fix the issue and move on without trying to context switch by discussing stuff.

PR #254, opened 2023-12-30
update jblow-nostalgia theme

jmpstar:
Made an attempt to update the theme to also cover welcome screen, footer and the different dialogs/panels. Let me know if there is more I missed.


focus-editor:
Looks way better now!

PR #256, opened 2023-12-31
save_current_buffer_on_build and save_all_file_backed_buffers_on_build settings

jlami:
So you don't forget to save your changes before building.
Save all files is limited to files that already have been saved once, so you don't get a save-as 'popup'.
I disabled save all by default and enabled save current buffer. We might want different defaults?
Settings are now 2 different bools, they should perhaps be set with one enum. Especially if we want to include saving new files as an option making it 3 bools instead of 1 enum. I could implement enum parsing in `parse_settings_line`

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -72,6 +72,8 @@ build_panel_line_wrap_always_on:        true
 build_panel_width_percent:              50
 build_panel_height_percent:             50
 
+save_all_file_backed_buffers_on_build:  false
+save_current_buffer_on_build:           true
 
 # This is an example configuration for build commands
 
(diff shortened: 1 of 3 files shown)


config/default.focus-config
```diff
@@ -72,6 +72,8 @@ build_panel_line_wrap_always_on:        true
 build_panel_width_percent:              50
 build_panel_height_percent:             50
 
+save_all_file_backed_buffers_on_build:  false
+save_current_buffer_on_build:           true
```
focus-editor, line 76: I will merge this but will make it false by default.

PR #257, opened 2023-12-31
Go to next build error

jlami:
Simple build error parser. Uses `error_regex` and `auto_jump_to_error` in Build_Command.

To support different compilers that might give different order for message, error type and/or line/column I added support for named captures in the `uniform` module. `map_named_captures` returns a Table that maps names to captures.

I use the regex `^(?P<file>.*):(?P<line>\d+),(?P<col>\d+): (?P<type>Error|Warning|Info): (?P<msg>.*)|^(?P<msg>.*error LNK.*)` in my own jai project.

There is a new action `go_to_next_build_error` that I mapped to `Ctrl-E` for myself. For the default we should probably poll Discord.

The action opens the next error in the active editor. If the next error in the queue is an info message (as is often the case in jai) it is opened on the side.

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -79,6 +79,8 @@ build_panel_height_percent:             50
 # build_working_dir:      <build working dir for all commands>
 # open_panel_on_build:    true  #  <- any settings specified here will apply to all commands unless overridden
 # close_panel_on_success: false
+# error_regex:            ^(?P<file>.*):(?P<line>\d+),(?P<col>\d+): (?P<type>Error|Warning|Info): (?P<msg>.*)|^(?P<msg>.*error LNK.*)
+# auto_jump_to_error:     true
 
 # [Debug Build And Run]   # <- command name. Can be arbitrary
 # build_command:          jai main.jai  # should be an executable or a script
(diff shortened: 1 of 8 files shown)


focus-editor:
Thanks for the PR. I will have to process the old PRs first which have been in the queue for a while, and I would appreciate if you changed your PR to match the style of the surrounding code in the mean time.

jlami:
Sorry, didn't see your request. Tried to clean up a bit. Thought the style matched pretty good already, but if you have any concerns please let me know.

focus-editor:
Hi, I've checked the PR and tested it. While it works, I'm not 100% happy with the behaviour and would structure the code differently. I will try to implement what I had originally planned using your code as a reference. It's easier than trying to shoehorn someone else's code into something else.

jlami:
Thanks for testing and let me know if I can help with anything.

focus-editor:
I ended up adopting a lot of what you wrote here @jlami and ditched my original idea of how this should be handled. Will be done soon. Thanks for your work!

focus-editor:
This can finally be closed. Thanks again!

PR #258, opened 2023-12-31
Make compatible with the latest jai beta (083)

valignatev:
Support for the double comma calling convention!

### DIFF
--- a/first.jai
+++ b/first.jai
@@ -12,7 +12,7 @@ RELEASE_DATE :: #run -> string {
 
 build :: () {
     options := get_build_options();
-    options.minimum_macos_version = .{12, 0};
+    options.minimum_os_version = .{12, 0};
     args := options.compile_time_command_line;
 
     import_path: [..] string;
--- a/modules/Input/examples/input_test.jai
+++ b/modules/Input/examples/input_test.jai
@@ -110,7 +110,7 @@ start_frame :: () {
             break;
         }
 
-        now := get_time();
+        now := seconds_since_init();
         if last_time dt = cast(float) (now - last_time);
         last_time = now;
         
--- a/modules/Input/windows.jai
+++ b/modules/Input/windows.jai
@@ -397,11 +397,11 @@ MyWindowProc :: (hwnd: HWND, message: u32,
  	 	        file_name_wide_count := DragQueryFileW(drop, i, null, 0) + 2;
  	 	        assert(file_name_wide_count > 0);
 
- 	  	        file_name_wide := cast(*u16) alloc(file_name_wide_count * size_of(u16));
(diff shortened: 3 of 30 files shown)


focus-editor:
That beta also changed `split` to not use the temp allocator by default, I believe, so we'll be leaking a lot unless we fix that too.

valignatev:
oh you're right! Fixed it too

focus-editor:
Thanks. I will have to attend to the other PRs first before merging this

ileonte:
`080` isn't even the latest compiler, smh my head.

valignatev:
yeah no worries, I'll rebase/fix conflicts when the time comes o/

Sanian-Creations:
This should not be merged until we either pull `modules/File_Async` into the repo ourselves, or the next beta fixes a problem with that module that causes it to not compile.
_beta-1-081_ fails to compile with `Error: Undeclared identifier 'temp'` for `modules/File_Async/io_completion_ports.jai`. It has some calls to `utf8_to_wide_new` with `,, temp`, but in that file Basic is a named import so it should be `,, Basic.temp`.

valignatev:
We can merge it to bump support from 79 to 80, but might as well wait  for 82 if it comes out soon

Sanian-Creations:
Also for 81:
* `get_time` got deprecated in favor of `seconds_since_init`
* `init_time`  got removed (probably should have been deprecated just like get_time, I think that's a mistake since it wasn't mentioned in the change notes)

valignatev:
Updated everything other than `init_time` calls. See discussion in Secret Beta discord. TLDR we can:
1) Change working with all times to `Apollo_Time`, it'll give way bigger precision
2) For now, export new time_init, that has `#file_scope` in Basic and call that. Resetting like that probably has some thread-safety implications but not sure how important it is

valignatev:
ok, it compiles and works with beta 83 now. Also rebased on top of latest main.

I've rewritten the code to use `Apollo_Time` for timestamps but keep the math and deltas to float64, seems like a nice middleground so we don't have 128bit things where we don't need them. Conveniently, we now use actual timestamps (as opposed to time since application start), so `current_time` is now an actual current time.

The only thing I haven't checked working is doubleshift (because it doesn't work on linux in general). So let me know if you see anything weird.

Sanian-Creations:
I think this is still missing some of the windows specific files.

valignatev:
Kk it compiles and works on windows now! Macos will require someone else to look at it though because I don't have have an easy way to set up macos environment

ileonte:
Beta `082` introduced this change:

> Renamed `Build_Options.minimum_macos_version` to `minimum_os_version` because it’s now also used to determine the minimum iOS version when cross-compiling for that target.

We should probably fix that in this PR as well.

focus-editor:
Thanks a lot

PR #267, opened 2024-01-01
Add `initializer_of` keyword and `Type` type for Jai

SentientCoffee:
- Reworked retroactive proc highlighting (based on Odin's proc highlighting)
  - Added exceptions so that most instances of `Poly_Type(T)` don't get mistakenly highlighted as procs
- Added `initializer_of` as a keyword
- Added `Type` type

### DIFF
--- a/src/langs/jai.jai
+++ b/src/langs/jai.jai
@@ -683,14 +683,14 @@ OPERATIONS :: string.[
 ];
 
 KEYWORDS :: string.[
-    "break", "case", "cast", "code_of", "continue", "defer", "else", "enum", "enum_flags", "for",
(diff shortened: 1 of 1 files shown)


focus-editor:
Sorry, I'm not sure I want to merge this change. I've already rejected a similar change which tried to improve type highlighting - #106 . The reason is that there's a line we must draw when doing this type of highlighting, after which we don't want to spend any effort improving the highlighting, because it gets progressively more expensive and still doesn't fix everything.

Given how often we use this tokenizer I don't want to make it slower just to fix some Type(T)  instances, which TBH are fine to be highlighted as procs because they could be thought as ones.

I would merge the rest of the changes if you modified the PR

SentientCoffee:
Understandable, I've reverted the major highlighting change.

PR #287, opened 2024-01-25
Jai: baked functions declaration highlighting

Roman-Skabin:
Just a suggestion for highlighting baked functions declarations


focus-editor:
Will do as an additional condition instead

PR #292, opened 2024-01-28
Added #library and #system_library modifiers highlighting

Roman-Skabin:
Actually it does not work perfectly when you use > 1 modifier: e.g. `#library,system,no_dll`. `system` will be highlighted (because we know that before_prev token was `#library`), but `no_dll` won't.


focus-editor:
I didn't realise there could be more than one modifier

PR #315, opened 2024-03-01
Added basic syntax highlighting for Uxntal

LainLayer:
image

Added basic syntax highlighting to the [Tal language](https://wiki.xxiivv.com/site/uxntal.html) for the Uxn virtual machine

I hope it works, I didn't experience or see anything particularly wrong on the files I tested with. If someone finds an issue, I would gladly fix it.

### DIFF
--- a/src/langs/uxntal.jai
+++ b/src/langs/uxntal.jai
@@ -1,0 +1,326 @@
+highlight_uxntal_syntax :: (using buffer: *Buffer) {
+    tokenizer := get_tokenizer(buffer);
+
+    while true {
+        token := get_next_token(*tokenizer);
+        if token.type == .eof then break;
+        prev := tokenizer.last_token;
+
+        color := COLOR_MAP[token.type];
+        memset(colors.data + token.start, xx color, token.len);
+
+        tokenizer.last_token = token;
+    }
+}
+
+#scope_file
+
+get_tokenizer :: (using buffer: Buffer) -> Uxntal_Tokenizer {
+    tokenizer: Uxntal_Tokenizer;
+
+    tokenizer.buf   = cast(string) bytes;
+    tokenizer.max_t = bytes.data + bytes.count;
+    tokenizer.t     = bytes.data;
+
+    return tokenizer;
+}
+
+get_next_token :: (using tokenizer: *Uxntal_Tokenizer) -> Token {
+    eat_white_space(tokenizer);
+
+    token := Token.{
+        start = cast(s32) (t - buf.data),
+        type = .eof
+    };
+
+    if t >= max_t then return token;
+    start_t = t;
+
+    char := t.*;
+
+                        /* I dont know what identifiers may start with */
+    if is_alnum(char) || char == #char "<" {
+        parse_identifier(tokenizer, *token);
+    } else if char == {
+        case #char ","; #through;
+        case #char "."; #through;
+        case #char ";"; #through;
+        case #char "_"; #through;
+        case #char "-"; #through;
+        case #char "="; #through;
+        case #char "&"; #through;
+        case #char "@";
+            parse_label(tokenizer, *token);
+
+        case #char "!"; token.type = .punctuation; token.punctuation = .jmi;          t += 1;
+        case #char "?"; token.type = .punctuation; token.punctuation = .jci;          t += 1;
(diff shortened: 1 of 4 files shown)


src/langs/uxntal.jai
```diff
@@ -0,0 +1,326 @@
...
+        t += 1;
+    }
+
+    if counter == 2 || counter == 4
+        token.type = .number;
+    else
+        token.type = .invalid;
+}
+
+is_hex_digit :: inline (char: u8) -> bool {
```
focus-editor, line 302: Was the `is_hex` function from `langs/common.jai` not good for these purposes?

In any case, why not use a couple of range checks instead of a case statement for every case?

src/langs/uxntal.jai
```diff
@@ -0,0 +1,326 @@
...
+    LDZ2r;  STZ2r;  LDR2r;  STR2r;  LDA2r;  STA2r;  DEI2r;  DEO2r;  ADD2r;  SUB2r;  MUL2r;  DIV2r;  AND2r;  ORA2r;  EOR2r;  SFT2r;
+    LIT;    INCk;   POPk;   NIPk;   SWPk;   ROTk;   DUPk;   OVRk;   EQUk;   NEQk;   GTHk;   LTHk;   JMPk;   JCNk;   JSRk;   STHk;
+    LDZk;   STZk;   LDRk;   STRk;   LDAk;   STAk;   DEIk;   DEOk;   ADDk;   SUBk;   MULk;   DIVk;   ANDk;   ORAk;   EORk;   SFTk;
+    LIT2;   INC2k;  POP2k;  NIP2k;  SWP2k;  ROT2k;  DUP2k;  OVR2k;  EQU2k;  NEQ2k;  GTH2k;  LTH2k;  JMP2k;  JCN2k;  JSR2k;  STH2k;
+    LDZ2k;  STZ2k;  LDR2k;  STR2k;  LDA2k;  STA2k;  DEI2k;  DEO2k;  ADD2k;  SUB2k;  MUL2k;  DIV2k;  AND2k;  ORA2k;  EOR2k;  SFT2k;
+    LITr;   INCkr;  POPkr;  NIPkr;  SWPkr;  ROTkr;  DUPkr;  OVRkr;  EQUkr;  NEQkr;  GTHkr;  LTHkr;  JMPkr;  JCNkr;  JSRkr;  STHkr;
+    LDZkr;  STZkr;  LDRkr;  STRkr;  LDAkr;  STAkr;  DEIkr;  DEOkr;  ADDkr;  SUBkr;  MULkr;  DIVkr;  ANDkr;  ORAkr;  EORkr;  SFTkr;
+    LIT2r;  INC2kr; POP2kr; NIP2kr; SWP2kr; ROT2kr; DUP2kr; OVR2kr; EQU2kr; NEQ2kr; GTH2kr; LTH2kr; JMP2kr; JCN2kr; JSR2kr; STH2kr;
+    LDZ2kr; STZ2kr; LDR2kr; STR2kr; LDA2kr; STA2kr; DEI2kr; DEO2kr; ADD2kr; SUB2kr; MUL2kr; DIV2kr; AND2kr; ORA2kr; EOR2kr; SFT2kr;
+}
```
focus-editor, line 106: Should really use metaprogramming to generate this enum - see examples in e.g. the jai tokenizer.

Some other languages are not using it for historical reasons - because we were hitting compiler bugs and had to remove the extra metaprogramming.

focus-editor (changes requested):
Thanks for the PR. You're likely to be the only one to use this language with this editor for quite a while :)

src/langs/uxntal.jai
```diff
@@ -0,0 +1,323 @@
...
+    for OPCODES print_to_builder(*b, "    %;\n", it);
+    print_to_builder(*b, "}\n");
+
+    return builder_to_string(*b);
+}
+
+Uxntal_Tokenizer :: struct {
+    using #as base: Tokenizer;
+    last_token: Token;
+}
```
focus-editor, line 124: Just noticed - you're using a special tokenizer but the only thing that's different from the default one is `last_token`, which is not used. Why not use the default one?

LainLayer: Oh you're right. I copied over the code and then realized I didn't need `last_token` and forgot about it. I'll change it.

PR #317, opened 2024-03-09
Selection length now factors in text length of additional cursors. Fix for the off-screen text overlap.

CELLTH:
Screenshot 2024-03-09 230109
It fixes this overlap as well.

Testing required.


focus-editor:
Thanks!

PR #319, opened 2024-03-14
Implemented command to open a new focus window

LainLayer:
The command duplicates the same command-line arguments used to launch focus the first time. It starts a new process independent of the first one.

https://github.com/focus-editor/focus/assets/10339438/56d64d75-3dba-44b7-8166-c7417097f2b7

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -196,7 +196,7 @@ Alt-A                       align_cursors
 Alt-Z                       toggle_line_wrap
 Alt-Shift-L                 toggle_line_numbers
 
-Ctrl-Shift-Alt-N            open_another_window
+Ctrl-Shift-Alt-N            duplicate_session
 
 [open file dialog]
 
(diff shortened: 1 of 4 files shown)


ileonte:
Could we maybe call the command something other than `open_another_window`? Something like `duplicate_session` perhaps? If we ever decide to do multi-window in the same Focus instance (which was discussed before) we probably want to "reserve" command names with `window` in them.

focus-editor:
How about we call it `open_another_editor_instance`?

Also @LainLayer , I noticed you made a typo in the command name, left a print statement in your PR and didn't follow the style of the surrounding code. I'm not a code style purist, but paying attention to those things shows that you care.

LainLayer:
I addressed the issues, however, I could not notice where I didn't follow the style of surrounding code. Maybe I missed something obvious?

focus-editor:
`duplicate_session` sounds confusing to me. Unless you guys have a better suggestion I'll rename it to `open_another_editor_instance` since it tells you exactly what it will do, whereas a "session" is ambiguous.

Re: style - I was referring to the space between the `[]` and the type

PR #323, opened 2024-03-23
Initial JS highlighting

simonvallz:
Add initial support for JS highlighting.

I don't have access to the compiler so I have a hard time testing it but I think this might be a good starting point.

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -1169,6 +1169,7 @@ get_tokenize_function :: (lang: Buffer.Lang) -> Tokenize_Function {
         case .Focus_Theme;  return highlight_focus_config_syntax;
         case .Glsl;         return highlight_glsl_syntax;
         case .Golang;       return highlight_golang_syntax;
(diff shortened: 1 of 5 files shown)


focus-editor:
Thanks. Impressive you're doing this without compiler access.
We'll try not to take too long to review it despite being busy.

IRooc:
Could this highlighting also work for `.ts` files? I know typescript has extra stuff like generics, but at least it is something

simonvallz:
I don't think it will be perfect with TypeScript (its probably not perfect for JavaScript either) but it will probably be better than no highlighting at all.

focus-editor:
@simonvallz thanks again for your work. I've fixed the errors and made the code a bit simpler. Here's a patch file, please apply it to your code and I will accept the PR.
[js-changes.patch](https://github.com/focus-editor/focus/files/14796237/js-changes.patch)

focus-editor:
Actually I just realised I can push the changes myself

PR #328, opened 2024-03-29
fixed segfault in c highlighting

LainLayer:
previous to this, starting with an empty C file and typing "#" would segfault.


focus-editor:
For some reason I couldn't reproduce it by doing what you said, but the change makes sense

LainLayer:
I had trouble reproducing it too. For some reason it did not always happen, but on some builds when I was testing it did. I couldn't verify what the actual variable was or if it was random. But it did happen to me, and the stack trace lead me to that spot

PR #345, opened 2024-04-18
Auto skip closing brackets

tebtro:
Similar to "auto_closing_brackets", were the closing bracket gets written when you type an open bracket.

Added "auto_skip_closing_brackets" to the settings config.
With this option enabled the closing bracket won't be written again if there already is one beneath the cursor.
The cursor will just be moved over by one position.


focus-editor:
Hi. Sorry, in the base editor we're not going to go in that direction. Adding auto-closing brackets was about as far as I was willing to go.

PR #362, opened 2024-05-08
Treat .hh file as C

CELLTH:
Similar to a .cc file


focus-editor:
Sorry, there's a major code revamp underway. I will merge it as soon as it's done

PR #364, opened 2024-05-14
Add "close_all_editors_and_buffers", and "close_other_buffers" commands.

maxamundsen:
The ctrl-tab menu tends to get clogged up as you open many files during an editing session. To remedy this, I created commands to clear the buffer menu. The "close_all_editors_and_buffers" closes everything, including the editors you are working on. The second command, "close_other_buffers", will clear the buffer list of every entry except the buffers currently visible in the editor(s).

**Commands added:**
close_all_editors_and_buffers
close_other_buffers


focus-editor:
Congrats on getting access to Jai!

There are quite a few issues with this PR. Did you look at it with a critical eye before submitting?

Do you want to find the issues yourself or do you want me to point them out to you?

maxamundsen:
I'll have to look at again tomorrow. I had a feeling something wasn't right, but it "works" in my testing. My first guess would be related to checking the editor state, and making assumptions about it?

focus-editor:
Yes. I suggest that you look at the surrounding code and make sure things are consistent with how it's done there. Command placement should not be in `handle_common_editor_action`. Function naming and command names should be consistent with the similar commands and with each other. The code itself can be simplified. There's no need to check for editor layout in the beginning - if we want to close all editors (or even other editors) we might still want to do it even if we don't have anything open.

maxamundsen:
I've removed the initial layout checks, and moved the commands into editors. For the naming, in the case of `close_all_editors_and_buffers`, naming it "Close All Files", seemed aligned with naming `close_current_editor` "Close File". This seemed logical to me. Is there a name that you would prefer? 

Regarding the code: My though process for actually closing "other" buffers was like this: You first need to clear the recent buffer array, then add only the buffers that are visible on the screen. If both buffers visible are the same, only add the active one. If there is a way to do this that leverages something already built in that I missed, please let me know!

src/editors.jai
```diff
@@ -853,6 +857,35 @@ do_things_on_editor_close :: (editor_id: s64) {
...
+    if editors.layout == {
+        case .Single;
+            array_add(*most_recent_buffers, editor_active.buffer_id);
+        case .Double;
+            editor_left := *open_editors[editors.left];            
+            editor_right := *open_editors[editors.right];
+            buffer_left := *open_buffers[editor_left.buffer_id];
+            buffer_right := *open_buffers[editor_right.buffer_id];
+            
+            if buffer_left == buffer_right {
```
focus-editor, line 874: 1. There's no need to compare the buffer pointers. Buffer ids can be compared just as well!
2. We don't even need to compare anything ourselves here, because all we need to do is to call the function `array_add_if_unique` twice.
3. There's currently a bug in this code. Because `editor_active` could point to the build panel and we never want to add it to the most recent buffers list.

src/editors.jai
```diff
@@ -853,6 +857,35 @@ do_things_on_editor_close :: (editor_id: s64) {
     }
 }
 
+close_other_buffers :: () {
+    array_reset(*most_recent_buffers);
```
focus-editor, line 861: Everywhere else we do `array_reset_keeping_memory` because we're likely to reuse that memory afterwards. Remember I suggested to look at the surrounding code and try to do the same thing.

src/widgets/commands.jai
```diff
@@ -119,6 +119,8 @@ base_commands := #run -> [] Command {
 
         cmd( "Go To Line",                        .go_to_line,                       .Single ),
 
+        cmd( "Close All Files",                   .close_all_editors_and_buffers,    .None   ),
+        cmd( "Close Other Buffers",               .close_other_buffers,              .None   ),
```
focus-editor, line 123: The command should be called Close Other Files (for consistency with the other commands).

focus-editor (changes requested):

src/widgets/commands.jai
focus-editor: Close Other Files should require the layout to be at least `.Single` (currently it requires `.None`)

src/editors.jai
```diff
@@ -110,6 +110,8 @@ active_editor_handle_event :: (editor: *Editor, buffer: *Buffer, event: Input.Ev
             case .move_up_fast;                     search_bar_move_cursor(editor, buffer, -5);                                         return true;
             case .move_down_fast;                   search_bar_move_cursor(editor, buffer,  5);                                         return true;
 
+            case .close_all_editors_and_buffers;    close_search_bar(editor); close_all_editors_and_buffers();                                                    return true;
+            case .close_other_buffers;              close_search_bar(editor); close_other_buffers();                                                              return true;
```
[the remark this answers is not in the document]
maxamundsen: Good to know. I was a bit confused with the terminology. Editor means an editor on screen, where a buffer is any opened file (that may or may not be on screen). By naming the command close_other_buffers, my intent was that you are closing other opened files, as the only "editors" were the buffers you are currently viewing on screen. Thank you for the clarification.

src/editors.jai
maxamundsen: My apologies for this one. Should've caught on to that.

src/editors.jai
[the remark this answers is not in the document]
focus-editor: An editor is an instance of an opened buffer with the corresponding cursors, scroll position, search results etc. You can have up to 2 editors pointing to the same buffer with the current design (limited to 2 panes).

maxamundsen:
Okay, I simplified the code, fixed the layout requirements, and made changes to the naming. Both commands are tested and working after the changes. 

After some testing, I noticed when the editor layout is single, both the left and right editors share the same buffer id. This removes the need to check the layout at all, since running `array_add_if_unique` twice will only add multiple buffers if two different editors are open. I suppose if we wanted to avoid the additional call to `array_add_if_unique` for the single layout, we could use an if statement... Not sure what your preference would be here.

src/editors.jai
```diff
@@ -857,32 +857,19 @@ do_things_on_editor_close :: (editor_id: s64) {
...
-            } else {
-                array_add(*most_recent_buffers, editor_left.buffer_id);
-                array_add(*most_recent_buffers, editor_right.buffer_id);            
-            }
-    }
+close_other_editors :: () {
+    array_reset_keeping_memory(*most_recent_buffers);
+
+    left_buffer_id := open_editors[editors.left].buffer_id;
+    right_buffer_id := open_editors[editors.right].buffer_id;
```
focus-editor, line 864: If you bind the `close_other_editors` command to a key combo and attempt to use it right away after launching the editor I bet it will crash.

focus-editor (changes requested):
Now it's getting closer to what it should be 👍

src/widgets/commands.jai
```diff
@@ -119,6 +119,8 @@ base_commands := #run -> [] Command {
 
         cmd( "Go To Line",                        .go_to_line,                       .Single ),
 
+        cmd( "Close All Files",                   .close_all_editors,                .Single ),
+        cmd( "Close Other Editors",               .close_other_editors,              .Single ),
```
focus-editor, line 123: Just noticed this is still called `Close Other Editors` when it should be `Close Other Files`.

src/editors.jai
maxamundsen: I tried this while debugging, and the function never gets called. Perhaps because it is an editor command, not a common command? Regardless, we can add a bounds check in `close_other_editors` to make sure editors.left and editors.right are not greater than the count of the array before indexing.

src/editors.jai
```diff
@@ -853,6 +857,26 @@ do_things_on_editor_close :: (editor_id: s64) {
     }
 }
 
+close_other_editors :: () {
+    if editors.left > open_editors.count || editors.right > open_editors.count {
+        return;
+    }
```
focus-editor, line 863: 2 issues with this:

1. This bounds check doesn't actually check bounds correctly, because it will still crash if e.g. `editors.left` = `open_editors.count`
2. The editor will still crash if this function is called before any editor is opened, because `editors.left` etc are initialised to `-1`

By the way, in this codebase we never put return on a separate line like this, instead it would go on the same line always.

focus-editor: Also, what if e.g. `editors.left` is out of bounds for some reason, but `editors.right` is valid - we don't want to just return and do nothing in that case

maxamundsen:
Hopefully the latest commit fixes those issues. editors.left and editors.right are checked independently before adding to `most_recent_buffers`. I double checked the conditions, and the bounds are now properly checked.

I know there is no style guide mentioned in the repo, but when you find it appropriate to utilize a `using` statement? As I'm learning more about this language, it seems super useful, but I do not want to abuse it. It seems to be used quite often reading other code for the editor, but I wasn't sure if you have a set rule of thumb.

I also apologize for all the silly mistakes I've been making, I've been working on this after programming at my job for 8+ hours, and my mental focus and judgement is sometimes diminished. I appreciate all the feedback you've been giving, and I'm impressed with your patience as I learn this codebase, and this language.

src/editors.jai
```diff
@@ -853,6 +857,33 @@ do_things_on_editor_close :: (editor_id: s64) {
...
+            left_buffer_id := open_editors[left].buffer_id;
+            array_add_if_unique(*most_recent_buffers, left_buffer_id);
+        }
+
+        if right_valid {
+            right_buffer_id := open_editors[right].buffer_id;
+            array_add_if_unique(*most_recent_buffers, right_buffer_id);
+        }
+    }
+}
```
focus-editor, line 879:
```suggestion
close_other_editors :: () {
    array_reset_keeping_memory(*most_recent_buffers);

    using editors;
    left_valid  := left >= 0  && left < open_editors.count;
    right_valid := right >= 0 && right < open_editors.count;
    if left_valid  then array_add_if_unique(*most_recent_buffers, open_editors[left].buffer_id);
    if right_valid then array_add_if_unique(*most_recent_buffers, open_editors[right].buffer_id);
}
```

We always want to reset most recent buffers, even if editor ids are out of bounds for any reason. Also made the code more compact and in my opinion easier to follow.

focus-editor:
It's OK, it's your first PR. I do appreciate it when people take the time and actually polish their PRs before submitting. In the past it's often not been the case unfortunately. People rush their PRs without paying attention to the style of the surrounding code and that makes me sad. It's a lot of work to review PRs and quite often it would be faster to just type the code myself, so any help with that from the contributors is appreciated.

As for the using of `using`, yes as long as you don't abuse it it's fine.

PR #370, opened 2024-05-24
Add Choose Language Dialog

maxamundsen:
Adds `choose_language` command and dialog for selecting the language of the current buffer. 

- The language strings in the dialog are generated at compile time from the `Buffer.Lang` enum. Perhaps we want a 'prettier' string to print, but this is a simple solution that doesn't require manually mapping a string for each enum.
- Once the dialog is open, the current language is pushed to the top, and it will show some text indicating that it is selected.
- If the build window is the active editor, the buffer language will not be changed, and a warning will pop up, alerting the user that you can't change the language of the build panel.
- When selecting `Plain_Text`, it will clear all of the active regions and tokens for the buffer, clearing all highlighting. (Originally, I implemented a basic tokenizer which set all tokens to .default, later realizing this was totally unnecessary 😂)

### DIFF
--- a/FOCUS-CHANGELOG.txt
+++ b/FOCUS-CHANGELOG.txt
@@ -23,6 +23,7 @@
         + Configs and themes will be migrated to version [4] with the new colors automatically derived from the old colors
         + You can always check the default config for the full list of available colors (use the `Show Default Config` command)
         + Do note that not all tokens are present in all languages. For example, `code_enum_variant` only works with Jai, Odin and Zig at the moment.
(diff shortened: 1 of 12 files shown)


focus-editor:
Overall I'm impressed with the quality of your PR. Thanks for your work.

I've made a few changes and am planning to make a few more later. This was actually the next item on my list and if you submitted it a few hours later there might've been conflicts. It might be a good idea to check with us before starting a sizeable piece of work to avoid conflicts or possible PR rejection.

PR #374, opened 2024-05-30
Added .comp as GLSL file type

St0wy:
`.comp` is often used for compute shaders in Vulkan, but right now it's not counted as GLSL.
This fixes that, I'm not sure if this could "collide" with HLSL (never used it, I don't know the conventions) but if not, would be a nice feature to have.


focus-editor:
If this causes problems in the future we have a solution in place in the next release - https://github.com/focus-editor/focus/blob/c507275def372bace869ea325f995ef178dbe7fa/FOCUS-CHANGELOG.txt#L28

PR #385, opened 2024-06-21
Separate data/config directories for Debug mode

cookednick:
When developing Focus, local config files (global, project, & themes) may be adjusted by test runs, potentially making them incompatible with the release version.

Now when DEBUG == true, Focus will create/use a "debug" subfolder *within* its usual data/config directories. Now, there is a separate set of data/configs for instances of Focus when compiled in Debug mode.

I made this change (and tested it successfully) on macOS, but I cannot test the Windows/Linux code paths. Please test this PR on those platforms before merging.


ileonte:
Looks good to me, honestly it's something I wanted to do for my use as well.

PR #388, opened 2024-07-17
fix sudden scroll to end of file when switching panels

### LINKED ISSUE #223: BUG: switching panel makes the view go to the bottom of the file
When switching panels the view goes all the way to the bottom of the file.

# Version
Focus 0.2.0, 26 november 2023, 21h01

Sanian-Creations:
Fixes #223

Moved two ifs down so that they perform their check on an updated version of `viewport.top` (which happens in the `if` on line 808, now 796), rather than the value from the previous frame.

Alternatively, it is possible to move the `if` on line 808 up to line 738 to update `viewport.top` sooner. That's is less code changes, but also has a bigger likelihood of changing the behaviour of other code in the file, though I've not noticed anything weird either way.

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -735,11 +735,6 @@ draw_editor :: (editor_id: s64, main_area: Rect, status_bar_height: float, ui_id
             cursors_start_blinking();
         }
 
-        // If we're past the max scrolling point for any reason (e.g. just deleted some text),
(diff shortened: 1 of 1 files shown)


focus-editor:
This code can be fragile but I'll test it and if I don't notice anything right away we can merge and keep an eye on it

PR #391, opened 2024-07-25
Fix comments toggling for all languages

Roman-Skabin:
Fixed toggle_comment and toggle_block_comment for already handled languages and added a pair of new ones which were not yet.
Also made it bit more robust.


Roman-Skabin:
Maybe show warning message for languages which do not support comments when you try to execute these commands?

focus-editor:
Very nice, thanks.
Yes, showing a short-lived warning message is definitely a good idea

PR #395, opened 2024-08-08
Basic CSS highlighting

simonvallz:
Basic CSS highlighting

### DIFF
--- a/src/langs/css.jai
+++ b/src/langs/css.jai
@@ -79,7 +79,7 @@ parse_identifier :: (using tokenizer: *Tokenizer, token: *Token) {
 
         t += 1;
         next_identifier_part := read_utf8_identifier_string(tokenizer);
-        identifier_str = sprint("%-%", identifier_str, next_identifier_part);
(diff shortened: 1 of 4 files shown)


src/langs/css.jai
```diff
@@ -0,0 +1,443 @@
...
+
+Property_Token :: struct {
+    type: Token_Type;
+    property: Property;
+}
+
+PROPERTY_MAP :: #run -> Table(string, Property_Token) {
+    table: Table(string, Property_Token);
+    size := PROPERTIES.count;
+    init(*table, size);
```
focus-editor, line 419: Might want to allocate some more here because hash tables like to have some free space to perform well. By default the target I believe is 70% fill rate, so this won't be enough and it will have to reallocate dynamically, and when it does the compiler sometimes crashes. At least that's what has happened to us before.
Maybe do `size * 2` here or something?

focus-editor:
Thanks for your work!

There are some invalid tokens in certain cases, if you have the time to handle them later please submit another PR.

image

PR #396, opened 2024-08-10
Parse raw_strings for Go

jejikeh:


src/langs/golang.jai
```diff
@@ -53,6 +53,8 @@ get_next_token :: (using tokenizer: *Tokenizer) -> Token {
         case #char "<";  parse_less_than             (tokenizer, *token);
         case #char ">";  parse_greater_than          (tokenizer, *token);
         case #char "!";  parse_bang                  (tokenizer, *token);
+
+        case #char "`";  #through;
```
focus-editor, line 57: I'd prefer it was a separate function for raw strings, because there's not much shared code.

src/langs/golang.jai
```diff
@@ -294,18 +295,30 @@ parse_bang :: (using tokenizer: *Tokenizer, token: *Token) {
...
         t += 1;
+        while t < max_t && t.* != #char "\n" {
+            if t.* == delimiter && !escape_seen break;
+            escape_seen = !escape_seen && t.* == #char "\\";
+            t += 1;
+        }
     }
-    if t >= max_t return;
 
+    if t >= max_t then return;
```
focus-editor, line 321: In this codebase we don't normally say `then return`, because `return` is itself a keyword visible enough to require an extra separation from the condition. A `then` is used when there's a normal statement after a condition, which would be harder to visually parse without a clear separator.

Any `then return`'s currently in the codebase got in there by accident

src/langs/golang.jai
```diff
@@ -293,6 +293,18 @@ parse_bang :: (using tokenizer: *Tokenizer, token: *Token) {
...
+
+    t += 1;
+    while t < max_t && << t != #char "`" {
+        t += 1;
+    }
+    if t >= max_t return;
+
+    t += 1;
+}
+
```
focus-editor, line 307: Much nicer!

PR #398, opened 2024-08-10
CSS highlighting improvements

simonvallz:
- Highlight keywords
- Only highlight units that are correct
- Increase table size

### DIFF
--- a/src/langs/css.jai
+++ b/src/langs/css.jai
@@ -33,7 +33,7 @@ get_next_token :: (using tokenizer: *Tokenizer) -> Token {
     if is_alpha(char) {
         parse_identifier(tokenizer, *token);
     } else if is_digit(char) {
-        parse_unit(tokenizer, *token);
+        parse_number(tokenizer, *token);
     } else if char == {
         case #char ".";  parse_simple_selector(tokenizer, *token);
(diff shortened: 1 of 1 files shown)


src/langs/css.jai
```diff
@@ -116,9 +123,18 @@ parse_unit :: (using tokenizer: *Tokenizer, token: *Token) {
...
-    while is_alpha(<<t) {
+    if <<t == #char "%" {
         t += 1;
-        if t >= max_t  return;
+        return;
+    }
+
+    temp_t := t;
+
+    unit_str := read_utf8_identifier_string(tokenizer);
```
focus-editor, line 133: I noticed that `read_utf8_identifier_string` is used everywhere in this file to read identifiers - is this really needed? I haven't been paying attention but I thought CSS didn't really support unicode. If that's changed then good, but still all the keywords are ASCII, so calling `read_identifier_string` would be cheaper in those cases

simonvallz: I think we can switch to `read_identifier_string` but perhaps the better solution would be a custom `read_css_identifier_string` that handles '-', '@' and '!' since a lot of keywords and properties have these?

focus-editor: Sounds like a good idea

PR #400, opened 2024-08-11
Fix an issue when toggle_block_comment is a fallback

Roman-Skabin:
If we fallback to `toggle_block_comment` we want to simulate `toggle_comment` behaviour: we want to comment out lines entirely, not just comment out a selection.


Roman-Skabin:
P.S. I've tested it on xml (it doesn't have single-line comments), works fine

PR #402, opened 2024-08-11
Fix a bug that caused a crash on Linux

simonvallz:
This change fixes a bug that made the editor crash on Linux when a file was passed as an argument. `editors_open_file` tried to set the title bar text before a window was created which resulted in a null dereference.

The bug was introduced in e4d9456d1d85958df25184e2451440d4d894b21c.


focus-editor:
Thanks for catching that

PR #416, opened 2024-09-01
Make the CSS parser simpler

simonvallz:


simonvallz:
Don't pretend we are a linter.

There are still some things that are not perfect but I think this is a step in the right direction.

focus-editor:
I agree it's not super important here to check for millions of keywords

PR #418, opened 2024-09-03
Add type keywords to JS/TS

simonvallz:


focus-editor:
Is it a keyword? I haven't been able to find any confirmation. It's also not highlighted in other editors

simonvallz:
It is a keyword in TypeScript but not in JavaScript. https://www.typescriptlang.org/docs/handbook/basic-types.html

But optimally it should only get highlighted when used as a type.

simonvallz:
Alternatively if the parser can figure out if things are types or identifiers so it can color all types in a special color I see no need to have it as a keyword. However, parsing TypeScript to figure this out does not seem to be a trivial task so it might not be worth it for now?

focus-editor:
If it's a type then we should use the TYPE_KEYWORDS list for that, even if we're not smart enough to detect if it's actually used as a type. Highlighting it as a general keyword would not be right.

focus-editor:
Guess we could add other basic types in there while we're at it

simonvallz:
Could also move boolean from KEYWORDS to TYPE_KEYWORDS if so.

simonvallz:
I added the type keyword I could find. Also added the `as` keyword.

focus-editor:
Cool, thanks!

PR #419, opened 2024-09-08
no_output build option

onelivesleft:
One style of developing is to compile frequently after making changes; either to jump to the next error you want to fix, or merely to check you haven't introduced one.  But, when you do this  on a working codebase you need to wait for the compilation to finish building, which feels long.  This PR lets you do this: `jai first.jai - no_output`
This will do the first stages of the compilation, checking for any errors in the code, but then exit as soon as it can.  On a working build this goes from 8+ seconds to 1.2 seconds on my PC.

### DIFF
--- a/first.jai
+++ b/first.jai
@@ -45,6 +45,8 @@ build :: () {
             options.null_pointer_check = .ON;
             options.arithmetic_overflow_check = .OFF;
           case "debug";
+          case "no_output";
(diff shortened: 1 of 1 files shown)


focus-editor:
8+ seconds for a debug build? This sounds very slow. It takes 3 seconds on my machine

PR #421, opened 2024-09-12
Fix alphabetical projects sorting order

Roman-Skabin:
Before:
image
After:
image

### DIFF
--- a/src/widgets/switch_to_project.jai
+++ b/src/widgets/switch_to_project.jai
@@ -46,11 +46,7 @@ switch_to_project_refresh_entries :: (filter: Fuzzy_Filter) {
     }
 
     // Load the list of most recent projects
-    if config.settings.projects_sorting_order == .most_recent_first {
-        for get_most_recent_projects() {
-            array_add(*project_configs, _entry_from_path(it));
-        }
-    }
+    for get_most_recent_projects()  array_add(*project_configs, _entry_from_path(it));
 
     // Load the list of projects from the global projects dir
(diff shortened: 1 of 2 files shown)


Roman-Skabin:
Oops, that's a bit incorrect. Right now we are trying to sort an empty array :)

Roman-Skabin:
most_recent_first:
image
alphabetical:
image
alphabetical_no_case:
image

src/widgets/switch_to_project.jai
```diff
@@ -46,11 +46,10 @@ switch_to_project_refresh_entries :: (filter: Fuzzy_Filter) {
...
     // Load the list of most recent projects
-    for get_most_recent_projects()  array_add(*project_configs, _entry_from_path(it));
-
-    if #complete config.settings.projects_sorting_order == {
-        case .most_recent_first;    // already sorted this way
-        case .alphabetical;         quick_sort(project_configs, (a, b) => compare_strings(a.name, b.name));
+    if config.settings.projects_sorting_order == .most_recent_first {
+        for get_most_recent_projects() {
+            array_add(*project_configs, _entry_from_path(it));
+        }
```
focus-editor, line 52: Why are we only doing this for `.most_recent_first`? This will not only change the sorting order but will also exclude any projects opened outside of the projects folder from the list.

Roman-Skabin: Hmm, I didn't think about projects which are not in focus' projects folder. I'll revert these lines

PR #422, opened 2024-09-15
Batch file support

arjenwitteveen:
I added syntax highlighting for Windows batch files. They're a bit of a nightmare to parse properly due to many context-dependent and obscure rules, but I think this does a decent job at handling the most common use cases.


focus-editor:
Very nice!

PR #423, opened 2024-09-19
Add subtle gradient effect to background

onelivesleft:


onelivesleft:
Added `gradient_effect` color setting; `effect_strength` is no longer hard-coded.

PR #424, opened 2024-09-19
CSS and TS improvements

simonvallz:
- Added `keyof` keyword to the JS parser.
- Fixed a bug that made the last character in a CSS comment not have the correct color.
- Added support for single line CSS comments. These are available in SCSS.


focus-editor:
I wonder if we should separate the js and ts parsers

PR #430, opened 2024-09-23
Add option to go to line in file when opening from command line

simonvallz:
E.g. `focus.exe first.jai:99`

### DIFF
--- a/src/main.jai
+++ b/src/main.jai
@@ -82,15 +82,37 @@ main :: () {
                 } else {
                     array_add(*dirs_from_args, dir);
                 }
-            } else if file_exists(path) {
-                editors_open_file(path);
             } else {
                 if path == "-project" {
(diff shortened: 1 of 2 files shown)


focus-editor (changes requested):

simonvallz:
Fixed the issues. Sorry for the inconvenience.

src/main.jai
```diff
@@ -82,15 +82,26 @@ main :: () {
...
+                    if found {
+                        line, success := parse_int(*line_str);
+                        if success && line >= 0 && file_exists(file) && !is_directory(file) {
+                            editors_open_file(file, line = line);
+                            opened = true;
+                        }
+                    } else {
+                        editors_open_file(file);
+                        opened = true;
+                    }
```
focus-editor, line 102: Why do we need this branch?

focus-editor: I see you deleted the code above, but why?

simonvallz: My bad, I didn't think this through. I removed this branch and re-added the one above.

PR #431, opened 2024-09-26
A couple of multicursor improvements you might like

onelivesleft:
Amended creating cursors above/below:
If the current cursor is not at the start of the line, when the user creates a cursor above or below it, empty lines will be skipped.  This is a lot more ergonomic; almost always this is what the user wants to do. The old behaviour is preserved when the cursor is at the start of the line (for instance, when on an empty line).

create_cursors_around keybind:
Cursors will be added to every contiguous line above and below the current cursor as long as the line in question reaches to the cursor column.


focus-editor:
Overall a very good change. 

Two things which bother me a bit:

1. When the cursor is on an empty line and you use `create_cursors_around`, it will add cursors to every line. I don't think we want that?
2. When there's more than one cursor on 2 different paragraphs. If you `create_cursors_around`, it will create cursors around the first cursor but not the second. I would expect it to either be disabled in multicursor mode or work on both cursors

onelivesleft:
1. Nope, will fix that.
2. Agreed, will fix that too!

focus-editor:
Now working perfectly, thanks!

PR #434, opened 2024-09-29
Added $extra_draw_info

onelivesleft:
extra_draw_info annotates draw calls; predominately when they are drawing the background. This doesn't currently change how they are drawn, but could in the future. Changed draw_rect_translucent_on_the_right to work via this mechanism.

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -112,7 +112,7 @@ draw_welcome_screen :: (main_area: Rect) {
 
     // Draw the left part
     {
-        draw_rect(left_area, Color.BACKGROUND_3);
+        draw_rect(left_area, .BACKGROUND_3, .{is_background = true});
 
         Simp.set_shader_for_rects();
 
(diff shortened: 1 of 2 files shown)


src/draw.jai
```diff
@@ -3532,9 +3534,17 @@ draw_top_down_shadow :: (rect: Rect) {
     draw_rect_raw(rect, c00 = t, c10 = t, c01 = s, c11 = s);
 }
 
-draw_rect :: (rect: Rect, color: Color) {
+draw_rect :: (rect: Rect, color: Color, $extra_draw_info := Simp.Extra_Draw_Info.{}) {
     Simp.set_shader_for_color(enable_blend = true);
-    draw_rect_raw(rect, color);
+    if extra_draw_info.is_translucent_on_the_right {
```
ileonte, line 3539: `#if` mayhap?

onelivesleft: Doesn't currently work, as the struct member doesn't count as constant. This is an optimisation that should be doable in future though, after a compiler improvement.

focus-editor: The compiler seems to not count as constants things that a clearly constants so that it almost feels like a regression because I thought it used to work before, but maybe I'm imagining things

PR #443, opened 2024-10-11
Switching buffers in the standard ctrl-tab way

onelivesleft:
Allows the user to switch between recent buffers, which should be an orthogonal operation to the cursor history (current `move_to_previous_buffer`/`move_to_next_buffer` use the cursor history).  This organises the buffers in order of recent access, per pane (i.e left / right editor).  While choosing a buffer, the filenames are drawn as tabs in the status bar.

Normally editors would bind this to ctrl-tab/ctrl-shift-tab; I've left the defaults for it as ctrl-pagedown/pageup for consistency with the current keymap, but would recommend rearranging these (and moving the current `switch_between_open_files` dialog to something else).  In this PR the ctrl key is hard-coded: once the user starts flipping through open buffers, we stay in this mode until the user releases ctrl.

Also added `close_current_buffer` (default Alt-w)

Other changes in this PR:
* Switched `most_recent_buffers` to our `Array` type.
* Renamed `Tween_Animation.speed` to `Tween_Animation.duration`
* `draw_file_info` now takes a `flags` parameter which governs which things to draw


onelivesleft:
This is usable as is, but needs a bit of polish on close_current_buffer when the buffer is modified, which I'm working on now

onelivesleft:
close_current_buffer polishing done.

focus-editor:
"Tabs" are now implemented in a branch as part of the editor system rewrite

PR #445, opened 2024-10-12
Add revert_select_word_or_create_another_cursor and move_selection_to_next_word commands

BuzzwordChief:
Hello,

my workflow involves heavy usage of the `select_word_or_create_another_cursor` command. To make this command more powerful this pull request adds two new commands:
1. `revert_select_word_or_create_another_cursor`: Reverts the action of `select_word_or_create_another_cursor`
2. `move_selection_to_next_word`: Executes a `select_word_or_create_another_cursor` and deletes the previously 'newest' cursor. It basically allows you to skip a selection.

For a better comprehension see the following demo video:

https://github.com/user-attachments/assets/7a590fa5-3bed-4e1f-825a-e92b9692238f

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -72,6 +72,26 @@ editors_handle_file_drop :: (files: [] string) {
     }
 }
 
+revert_select_word_or_create_another_cursor :: (using editor: *Editor, buffer: *Buffer) {
+    if cursors.count == 1 {
+        cursor := *editor.cursors[0];
+        cursor.pos -= cursor.pos - cursor.sel;
+        cursor.sel  = cursor.pos;
+        cursor_moved = .unimportant;
+    } else {
+        organise_cursors(editor);
+        
+        ordered_remove_by_index(*cursors, main_cursor);
+        if main_cursor == 0 {
+            main_cursor = cursors.count - 1;
+        } else {
+            main_cursor -= 1;
+        }
+
+        cursor_moved = .selection;
+    }
+}
+
 active_editor_handle_event :: (editor: *Editor, buffer: *Buffer, event: Input.Event, action: Action_Editors) -> handled: bool {
     context.current_editor_id = editors.active;  // used for edit notifications
     defer context.current_editor_id = -1;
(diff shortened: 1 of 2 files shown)


src/editors.jai
```diff
@@ -1792,6 +1795,71 @@ select_word_or_create_another_cursor :: (using editor: *Editor, buffer: Buffer)
     return editor.search_whole_words;
 }
 
+revert_select_word_or_create_another_cursor :: (using editor: *Editor, buffer: *Buffer) {
+    if !get_selected_text_all_cursors(editor, buffer) || cursors.count == 1 {
+        for *cursors {
+            it.pos -= it.pos - it.sel;
+            it.sel  = it.pos;
```
focus-editor, line 1802: If the purpose of this is to remove selection, then `cursor.pos = cursor.sel` should suffice.

After testing this functionality I think we shouldn't remove selection when there's more than one cursor and the selected text is different

src/editors.jai
```diff
@@ -1792,6 +1795,71 @@ select_word_or_create_another_cursor :: (using editor: *Editor, buffer: Buffer)
...
+        }
+    }
+    cursor_moved = .selection;
+    
+    assert(main_cursor     < cursors.count);
+    assert(original_cursor < cursors.count);
+}
+
+move_selection_to_next_word :: (using editor: *Editor, buffer: Buffer) 
+{
```
focus-editor, line 1832: In this codebase we don't put opening braces on a new line

src/editors.jai
```diff
@@ -1792,6 +1795,71 @@ select_word_or_create_another_cursor :: (using editor: *Editor, buffer: Buffer)
...
+            main_cursor -= 1;
+        }
+        
+        if original_cursor > old_main_cursor {
+            original_cursor -= 1;
+        } else if original_cursor == old_main_cursor {
+            // This is suboptimal but can happen. If select_word_or_create_another_cursor does not find a new
+            // word to select it will set the main_cursor to the original_cursor. So it is completely possible
+            // that we revert the original_cursor.
+            // Maybe select_word_or_create_another_cursor should do nothing if it can't find a new word to select?
```
focus-editor, line 1821: No, by wrapping the main cursor back to the original cursor we know we have selected everything, it is an intuitive behaviour which should be kept.

However, with this change it feels very unintuitive to remove the original cursor when this action is used after we've fully wrapped around. It should never remove the original cursor and it should always be possible to revert back to it by repeatedly using this action.

focus-editor (changes requested):
A very useful change, thanks for submitting it - please look at the requested changes

src/editors.jai
BuzzwordChief: Done, see line editor.jai:1800.

src/editors.jai
BuzzwordChief: Done

src/editors.jai
BuzzwordChief: Agreed. New version checks for original_cursor == main_cursor and deletes the cursor above the main_cursor in this case, see line editor.jai:1813.

focus-editor:
Seems to be working well now, thanks!

PR #452, opened 2024-10-25
Fixes for Move by word 

onelivesleft:
Fixed select_word on whitespace extending over newlines.
Fixed _ends skipping non-word chars too far.
Removed underscore_is_part_of_word from context.
Note: cleaned up char queries in utils.jai; any bugs wrt is_word_char in the near future is probably due to this.


onelivesleft:
Final form:
* move_#_by_character_type (Default with alt on pc)
Moves across one character group (i.e. word, number, space, other)

* move_#_through_word (Default with ctrl on pc)
Moves until it has moved through a `word` character group.

* move_#_through_word_throttled
As above, but will only go through at most a space + 2 character groups.

PR #457, opened 2024-11-03
Prevent out of bound array access in color preview

darkgiggs:
Previously, when changing region colors such as region_heredoc with the color preview opened, adding a colon as the last character on the last line would make the editor crash.

### DIFF
--- a/src/main.jai
+++ b/src/main.jai
@@ -75,7 +75,7 @@ main :: () {
         for path : array_view(args, 1, args.count - 1) {
             if is_directory(path) {
                 dir := trim_right(path, "\\/");
-                project_config_path := tprint("%.focus-config", dir);
+                project_config_path := tprint("%/.focus-config", dir);
                 if !project_config.loaded && file_exists(project_config_path) {
                     success := load_project_config(project_config_path);
                     if !success then add_user_error("Tried to load project config from % but couldn't due to errors", project_config_path);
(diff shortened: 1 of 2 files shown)


darkgiggs:
Regarding https://github.com/focus-editor/focus/pull/457/commits/7e1172ed74d9508ffbb8eba0821970862b3198d0
If I'm reading this right, the intent is to have project directories inside the "projects" directory, with each config file at the "projects" root?
I don't think there was a way to construct a valid config-file path with the previous code, but I hope this fix is the intended one.

focus-editor:
What was the problem you were having with config paths that led you to change that line?

focus-editor:
As this is unrelated to the crash fix, this makes it hard to merge. It should ideally have been 2 PRs

focus-editor:
I'll be happy to merge this, however I'd like to reproduce the crash first.

I've tried changing the `region_heredoc` color with the following test text:

```
HEREDOC :: #string LANG_C
#include <stdio.h>

int main(int argc, char **argv) {
    printf("Hello, world!\n");
    return 0;
}
:
```

I couldn't get it to crash wherever I put the colon. 

Please can you tell me what I'm doing wrong?

darkgiggs:
The crash happens as you type the colon for setting a color if that color is on the last line of the file.
It happened to me while attempting to edit a project config-file.

The easiest repro steps are probably:
- Open the global config file
- Delete everything under [colors]
- Have the preview window open on the Jai section
- Type a parameter and the colon

It crashes on current release for me, here's a screenshot

image

focus-editor:
Thanks for bringing attention to this problem. This code actually needs a rewrite as it has other problems besides the crash. I'll merge your fix for now before other fixes are done

PR #463, opened 2024-11-08
Don't copy to clipboard in delete_to_{start,end}_of_line

simonvallz:
The delete_to_{start,end}_of_line actions no longer copies the text to the clipboard. Two new actions cut_to_{start,end}_of_line has been added that does copy the text to the clipboard.


focus-editor:
Nice work!

Issue #454, opened 2024-10-28
[MacOS] Caret is offset vertically on second monitor


MrStevns:
**Describe the bug**
When I move a focus window onto a second monitor and start selecting something, the caret is offset on the a axis.

**To Reproduce**
1. Open Focus and start a file with multiple lines
2. Move the focus window to a second monitor
3. Select something

**Expected behavior**
Selection is started from cursor position

**Actual behavior**
Selection is being made somewhere above the actual content.

**Video**
https://github.com/user-attachments/assets/acfc391a-b0fa-41c0-9123-0d3461cdca50

 - OS: MacOS Monterey
 - Focus Version: 0.3.7

focus-editor:
Does it happen every time?

It doesn't reproduce on my mac, hmmm

MrStevns:
Yes it happens consistently.
I have tried to close and reopen focus too, it happens every time for me.

Ah another thing, my macbook is retina (2560x1600) while my second monitor is 1080p. Maybe the issue is caused by the difference in resolution.

focus-editor:
Latest nightly too?

focus-editor:
https://drive.google.com/file/d/1h1PO-Ft1_ad5-FekbvYOxr81qIt8aYvZ/view?usp=drive_link
here's a build

MrStevns:
Yep, In nightly too.

I however just discovered that when you put the application in fullscreen, the problem gets resolved for that monitor. When I move the window back to first monitor, the caret offset again but only for that monitor.

focus-editor:
I have tried it on my macOS Sequoia with different monitors and different display resolutions, can't reproduce. Could be an older macOS issue, but unfortunately I don't have a Monterey machine to test it on.

simonvallz:
I am able to reproduce this and it seems to happen when the window is moved to a display with a different backing scale factor. Here is a PR with a fix that at least works for me:  #467

focus-editor:
On my mac there seems to be no way to select a scaling factor for a display, but I played with different resolutions and wasn't able to reproduce it. 
Maybe they "fixed" it in newer macOS's, but your fix seems reasonable, so we'll merge it, thanks!

focus-editor:
@MrStevns please try this build https://drive.google.com/file/d/1Ue4L-zpOUs2Qc2WeN2MFuucIL-MCjV9F/view?usp=drive_link

simonvallz:
[...]

PR #467, opened 2024-11-13
macOS: Re-compute width and height when window changes display

simonvallz:
Different displays can have different backing scale factors so we need to re-compute the width and height of the window when it is moved to another screen.


simonvallz:
Please also see #454

PR #474, opened 2024-12-04
Macos title bar config options

maxamundsen:
With this change, the MacOS build now respects the title bar color options from the config.


focus-editor:
Working like a charm

PR #482, opened 2024-12-23
Add syntax highlighting in JS template strings

simonvallz:


src/langs/js.jai
```diff
@@ -423,6 +505,22 @@ parse_decorator :: (using tokenizer: *Tokenizer, token: *Token) {
     identifier_str := read_utf8_identifier_string(tokenizer);
 }
 
+at_string :: (using tokenizer: *Js_Tokenizer, a: string, $case_sensitive := true) -> bool {
+    if t + a.count > max_t  return false;
+    b: string = ---;
+    b.data = t;
+    b.count = a.count;
```
focus-editor, line 512:
```suggestion
    b := string.{ count = a.count, data = t };
```

simonvallz: I changed to this and moved `at_string` to `common.jai`.

PR #483, opened 2024-12-23
Add syntax highlighting for Swift string interpolation

simonvallz:


src/langs/swift.jai
```diff
@@ -433,6 +516,22 @@ parse_dollar :: (using tokenizer: *Tokenizer, token: *Token) {
...
 
+at_string :: (using tokenizer: *Swift_Tokenizer, a: string, $case_sensitive := true) -> bool {
+    if t + a.count > max_t  return false;
+    b: string = ---;
+    b.data = t;
+    b.count = a.count;
+    #if case_sensitive  return equal(a, b);
+    else                return equal_nocase(a, b);
+}
+
```
focus-editor, line 527: This (and similar functions) should be implemented on the generic `Tokenizer` and be in `common.jai`, since they only use the generic fields.

simonvallz: Moved this function to `common.jai`. Also updated `batch.jai` to use the version from `common.jai`. It seems like `xml.jai` also uses this but `Xml_Tokenizer` is not `#as using base: Tokenizer;` so I skipped that for now.

PR #484, opened 2024-12-23
Fix a bug in draw_range_full_width_lines if range is only one line

simonvallz:
This fixes a bug where the background color in a single line buffer region was too wide.

E.g. #string jai struct jai


focus-editor:
I noticed that when using JS <script> tags it still doesn't look as expected:
image

1. Black square even when there's nothing in between tags
2. Last character is not within the region when there is something in between tags
3. Highlights start at a weird place when tags are on different lines. Ideally we should only highlight the space in between the tags without the lines where the tags are (unless there's something there besides the script tags(

simonvallz:
I can have a look at this in a couple of days. :)

PR #486, opened 2024-12-24
Jai 0.2.002 hotfix

ebootdpr:
hold tag?


modules/Input/macos.jai
ebootdpr, line 1: I dont know if  it should be
jai_context: #Context;
or
jai_context: Context_Base;
idk

focus-editor:
1. We're not going to upgrade to Jai 0.2 until at least the next editor release, because it's potentially unstable
2. By replacing Runtime_Support with a newer version the very useful error reporting functionality is lost, we can't merge that.

focus-editor:
We've upgraded to 0.2.010 finally

PR #487, opened 2024-12-25
Improve HTML script background drawing

simonvallz:
The background coloring of buffer regions has been changed to follow these rules:
- Don't draw background for a buffer region only containing whitespace.
- Don't draw background on the first and last line if there is nothing but whitespace on those.
- Otherwise draw the background.

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -3324,6 +3324,8 @@ draw_range_full_width_lines :: (range: Coords_Range, editor_rect: Rect, visible_
     start := max(range.start.line, visible_start);
     end   := min(range.end.line,   visible_end);
 
+    if start == end && range.start.col == range.end.col  return;
+
     full_width_rect := Rect.{ editor_rect.x, text_origin.y - start * line_height, editor_rect.w, line_height };
     text_gap_x := text_origin.x - editor_rect.x;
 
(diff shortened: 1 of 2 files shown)


src/draw.jai
```diff
@@ -1267,10 +1267,32 @@ draw_editor :: (editor_id: s64, main_area: Rect, status_bar_height: float, ui_id
...
+                last_line  := offset_to_line(editor, buffer, end);
+
+                first_line_end  := get_line_end_offset(editor, buffer, first_line);
+                last_line_start := get_line_start_offset(editor, buffer, last_line);
+
+                first_line_text := get_range_as_string(buffer, .{ start = start, end = first_line_end });
+                last_line_text  := get_range_as_string(buffer, .{ start = last_line_start, end = end });
+
+                if is_all_whitespace(first_line_text)  start = first_line_end + 1;
+                if is_all_whitespace(last_line_text)   end   = last_line_start;
```
focus-editor, line 1288: This seems very expensive to do each frame.

We should instead adjust the region boundaries at the time when they are determined in the highlighter, not every frame before drawing.

simonvallz: I moved this into the HTML parser.

focus-editor:
Working very well now 👍

PR #492, opened 2024-12-31
Fix a string length calculation bug

simonvallz:
This fixes an issue that could be seen on the last line in a script tag if there is a single JS character just before the closing script tag:

<script>
	const a =
1</script>

In this case `1` didn't have the correct background color.


focus-editor:
@simonvallz still seems broken
image

PR #494, opened 2024-12-31
Add syntax highlighting for unix shell languages (sh/bash/zsh)

SentientCoffee:
Add support for Unix shell-based languages -- POSIX sh, Bash, and Zsh. The highlighting is set up for Zsh since it's a superset of the other two, so even normal .sh files get highlighted with some zsh specific things.


focus-editor:
Thanks!

PR #495, opened 2025-01-01
Make single line buffer regions starting from col 0 draw properly

simonvallz:
This fixes a bug that made single line buffer regions that start from column 0 not include `text_gap_x` in the width.


simonvallz:
This fixes https://github.com/focus-editor/focus/pull/492#issuecomment-2566745936

focus-editor:
Working now!

PR #496, opened 2025-01-02
Add tab navigation to Generic Confirm Dialog

caztanj:
Do you want this? I like it.

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -239,7 +239,6 @@ Ctrl-Shift-Alt-N            open_another_editor_instance
 
 [open file dialog]
 
-Enter                       open_entry_in_place
 Ctrl-Enter                  open_entry_on_the_side
 Ctrl-1                      open_entry_on_the_left
 Ctrl-2                      open_entry_on_the_right
@@ -252,7 +251,6 @@ Backspace                   pop_directory
 
 [search dialog]
 
-Enter                       open_entry_in_place
 Ctrl-Enter                  open_entry_on_the_side
 Ctrl-1                      open_entry_on_the_left
 Ctrl-2                      open_entry_on_the_right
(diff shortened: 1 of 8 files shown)


focus-editor:
It's looking good, I like it, however a few things need to be changed before we can merge this:

1. There's already an action for this - currently in the search dialog section. It needs to move to the common section, and there should be no need to add another action
2. This needs to work with the delete file dialog - actually we might need to make the delete dialog be a generic confirm dialog
3. We shouldn't check for .ENTER, instead there should be an action for this in `[common]`, I suggest moving the `open_entry_in_place` in there

focus-editor:
Just realised that the delete file dialog was already migrated to the generic confirm dialog, sorry, it's been a while and I forgot already

simonvallz:
I fixed point 1 and 3. Do I need to do something in config migrator when a default key bind is moved to `[common]`?

simonvallz:
Or maybe you want to have `open_entry_in_place` in all three places (`[common], [open file dialog], [search dialog]`)?

focus-editor:
Nice. 

> Do I need to do something in config migrator when a default key bind is moved to [common]

In our case we should just add it to the ADDED_KEYBINDS for the common section. When migrating user configs it will stay in their configs in the dialog sections, but that's ok because it will still work, but in the default config it should only be in the common section.

Please can you also rename the `go_to_next_input` action to something which works with both inputs and buttons since it's more generic. Maybe `focus_next_ui_element`?

simonvallz:
I added it to ADDED_KEYBINDS now and I think `focus_next_ui_element` sounds good so I changed it to that. Also updated RENAMED_KEYBINDS with the change.

focus-editor:
Thanks, looking good. Sorry I was travelling and was away from my PC, only back now.

PR #502, opened 2025-01-08
Fix crash on reloading a workspace if a file path is too long

Roman-Skabin:
I'm not sure whether it is the correct place to check it or not. But the issue comes from `find_or_create_buffer` returning -1 because of this:
image


Roman-Skabin:
Honestly, I would just rather make files work with long paths. That's not a rocket science, just prepend `\\?\` to the filename.
Or, if you don't want to prefix filenames, you can add `<longPathAware>true</longPathAware>` to the application manifest.

focus-editor:
I wonder if we should just say we support only Windows 10, version 1607 or higher.
Probably will cover 99.9% of Windows users.

PR #507, opened 2025-01-24
Added autosave in x seconds feature.

pmjames29:
Hey, I implemented a feature for autosaving modified files in x seconds. Currently, it works by checking if there is no input for the specified amount of time in the configuration, and then it will save all modified buffers in the editor. For files that have been modified on the disk with changes in the editor, it will not autosave the new changes. There is a bug currently when inputting backspaces, as it does not reset the timer for the autosave and will autosave as if no changes are being made, even though there is user input. The session modified buffer functionality has some overlap with this, and once it is implemented it may be able to fix the problem. You may already be trying to implement this with the session modified buffer code, so you might want to discard or take what pieces you want from this, thanks.

### DIFF
--- a/config/default.focus-config
+++ b/config/default.focus-config
@@ -61,6 +61,7 @@ prefer_system_file_dialogs:                         false  # Windows only
 persist_local_search_results:                       false  # if true, search results will stay highlighted and you have to dismiss them using the `escape` action
 load_most_recent_project_on_start:                  false
 projects_sorting_order:                             most_recent_first
+autosave_in_x_seconds:                              0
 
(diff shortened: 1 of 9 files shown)


obiwanus:
Thanks for your PR. 

We're pausing all PRs until the next release, we'll have a look after it's done.

I had a quick look, one thing I noticed - why do we need an additional timer? There's an auto-refresh timer already which should be sufficient.

focus-editor:
Thanks for waiting, sorry it took so long. 
Would you mind fixing the conflicts and resubmitting the PR, when we can merge it?

Note that the autosave timer isn't needed, we have a refresh timer already which ticks every second, and we don't want sub-second precision with the autosave feature.

PR #510, opened 2025-01-28
Attempt to make directory when saving as

caztanj:
If the user enters a filename like "test/dir/file" we create the test/dir directory before saving the file.


src/deferred_actions.jai
```diff
@@ -219,6 +219,8 @@ execute_save_as :: (action: Deferred_Action) -> Action_Result {
             push_action(.confirm_overwrite, .{ confirm_overwrite = .{ buffer_id = buffer_id, is_move = is_move, file_path = copy_string(file_path) } });
             return .NEXT;
         } else {
+            found, directory, file := split_from_right(file_path, #char "/");
+            if found && directory.count > 0  make_directory_if_it_does_not_exist(directory, recursive=true);
```
focus-editor, line 223: Would you be ok to add handling for the case when the directory wasn't able to be made please

focus-editor:
It does save directory, but typing a slash is blocked in that dialog - I'm going to have to have a look

Issue #397, opened 2024-08-10
Request: Add syntax highlighting for Java


HSGWHRSI:
Please spare me from having to use an IDE.

focus-editor:
Ideally someone familiar with Java would submit a PR for this. Even someone without access to jai could do it by copying the C highlighter and changing keywords. We could fix it afterwards to make sure it works

patrickbard:
> Ideally someone familiar with Java would submit a PR for this. Even someone without access to jai could do it by copying the C highlighter and changing keywords. We could fix it afterwards to make sure it works

That's exactly what I did 😀

ileonte:
Java highlighting is now available in the latest nightly builds.

PR #513, opened 2025-02-08
Added syntax highlighting for Java

### LINKED ISSUE #382: Feature Request: Add Java Language Support
Add Java language support.

patrickbard:
I'm opening this to add support for Java syntax highlighting. Fixes #382 and #397.

Since I don't yet have access to Jai, I did my best to implement this appropriately without being able to test it.
I started based on the `lang/c.jai` file and changed what was necessary.
I also compared this change with similar syntax pull requests.
But I guess more testing should be done by someone who can actually run and debug it.

In the hopes that it might be useful, I included some comments mentioning the current Java Specification document.
Another reason for it was to not implement this support based only on my experience with the language.

Better support for JavaDoc comments could be looked for in the future, since it's common to differentiate them from regular multiline comments. I think that would require a bigger change, and I was not feeling comfortable doing that without debugging.

### DIFF
--- a/src/buffer.jai
+++ b/src/buffer.jai
@@ -550,6 +550,7 @@ tokenize_for_indentation :: (buffer: *Buffer) -> [] Indentation_Token /* temp */
         case .Cpp;          return tokenize_c_like_lang_for_indentation(buffer, get_next_cpp_token);
         case .Css;          return tokenize_c_like_lang_for_indentation(buffer, get_next_css_token);
         case .D;            return tokenize_c_like_lang_for_indentation(buffer, get_next_d_token);
+        case .Java;         return tokenize_c_like_lang_for_indentation(buffer, get_next_java_token);
         case .Js;           return tokenize_js_for_indentation(buffer);
         case .Json;         return tokenize_c_like_lang_for_indentation(buffer, get_next_json_token);
         case .Glsl;         return tokenize_c_like_lang_for_indentation(buffer, get_next_glsl_token);
@@ -1600,6 +1601,7 @@ get_tokenize_function :: (lang: Buffer.Lang) -> Tokenize_Function {
         case .Glsl;                 return tokenize_glsl;
         case .Hlsl;                 return tokenize_hlsl;
         case .Golang;               return tokenize_golang;
+        case .Java;                 return tokenize_java;
         case .Js;                   return tokenize_js;
         case .Json;                 return tokenize_json;
         case .Lua;                  return tokenize_lua;
@@ -1839,6 +1841,7 @@ Buffer :: struct {
         Glsl;
         Hlsl;
         Golang;
+        Java;
         Js;
         Json;
         Lua;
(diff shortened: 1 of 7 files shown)


focus-editor:
Thanks for your PR.
We're pausing all PRs until a major piece of work is finished, then we'll
get back to them.

ileonte:
I've had a quick look at the PR and it's mostly good. There are a couple of small issues and a pretty big one.

The small issues:
* you forgot to rename `get_next_c_token` to `get_next_java_token` on line `31` of `java.jai`
* there's a missing `,` after `"yield"` on line `418` of `java.jai`

The big issue: you added the string `"non-sealed"` to the `MODIFIER_KEYWORD` array, which is not good for two reasons:
1. the strings from that array get translated to Jai enum members in the metaprogram. This results in the identifier `kw_non-sealed` being generated, which causes compilation to fail.
2. even if we could somehow work-around the Jai compilation issue our parsers cannot currently interpret `non-sealed` as a single token.

If you want I can merge the PR as is, fix the small syntax issues and work-around the `non-sealed` problem by removing it from the list of keyword modifiers.

focus-editor:
Guess we could replace - with _ when generating identifiers?

patrickbard:
> I've had a quick look at the PR and it's mostly good. There are a couple of small issues and a pretty big one.

Thank you for taking a look.

> The small issues:
> 
> * you forgot to rename `get_next_c_token` to `get_next_java_token` on line `31` of `java.jai`
> * there's a missing `,` after `"yield"` on line `418` of `java.jai`

Yeah, I thought that something like that could happen since I was relying only on my memory and diffing skills.

> The big issue: you added the string `"non-sealed"` to the `MODIFIER_KEYWORDS` array, which is not good for two reasons:
> 
> 1. the strings from that array get translated to Jai enum members in the metaprogram. This results in the identifier `kw_non-sealed` being generated, which causes compilation to fail.
> 2. even if we could somehow work-around the Jai compilation issue our parsers cannot currently interpret `non-sealed` as a single token.

This totally makes sense, but I guess this would be very difficult to realize since I never used Jai and could not compile and check for errors.

> If you want I can merge the PR as is, fix the small syntax issues and work-around the `non-sealed` problem by removing it from the list of keyword modifiers. Or maybe you can make the modifications on your branch and I can merge in one go. Let me know.

Let me fix it.
I can do this tomorrow.

patrickbard:
> > Guess we could replace - with _ when generating identifiers?
> 
> We could but then we would also highlight `non_sealed` as a keyword which is incorrect, it needs to be `non-sealed` and we don't support that at the moment.

I assume @focus-editor meant doing that only when constructing the enum members. It would then be necessary to perform some checks and change "_" back to "-" when parsing the identifiers, I guess.

I just don't know what the best way to do this in Jai for this project, taking performance into consideration.

It seems that it would require to add "-" to `read_utf8_identifier_string` but I don't know if that could have side effects. And it would probably require another check inside Java `parse_identifier` in order to account this special enum member case.

patrickbard:
Another approach to solve this could be, instead of dropping support for `non-sealed`, adding `non` as a separate keyword.

This would be wrong according to the specification and would incorrectly highlight a individual "non" as a keyword.

But, since there is already another keyword for `sealed`, it would display correctly (ignoring the dash) for users that have the `non-sealed` keyword in actual use.

patrickbard:
Thank you @ileonte for merging my PR and solving my mistakes.
I had a big internet outage in my region so i could not fix it myself as I said I would.

I love this project, I'll try to contribute more in the future.

PR #516, opened 2025-02-13
Add patch/diff file highlighting

SentientCoffee:
Added patch/diff file highlighting for easy at a glance viewing.

Does not try to highlight based on language, just a simple green for additions and red for deletions, plus some extra colours for metadata.

(This could be changed in the future if we did want to highlight per language since the regions I added could stay the same and just change the token colours... That would involve needing to detect language based on the file extensions of the diff'd files though which might be kind of slow, especially for large patches that involve many files.)


SentientCoffee:
Forgot to update the unit tests and themes, whoops.

obiwanus:
Thanks for the PR! Sorry @SentientCoffee, this patch will have to wait because if we merge it now we'll have to fix many conflicts with the long running branch which is blocking everything. In fact I wish you hadn't updated the config, tests and themes, because it would've been easier to add them later

SentientCoffee:
Is it just the default config/themes/tests that would cause an issue? I can revert those if needed, I updated them because otherwise `jai first.jai - release` fails the unit tests.

focus-editor:
Yes please, then it will be easier to merge later

SentientCoffee:
Reverted the unit test and theme changes 👍🏻

obiwanus:
Great, thanks. If you want a release build for yourself in the mean time, you can just disable tests temporarily.

focus-editor:
Hey @SentientCoffee - sorry it took so long to get back to your PR.
Would you mind resolving the conflicts so that we can merge it please?

SentientCoffee:
Conflicts resolved. Had to bump the config version again, but as far as I can tell everything works fine afterward.

src/langs/diff.jai
```diff
@@ -0,0 +1,145 @@
...
+}
+
+parse_plus :: (using tokenizer: *Diff_Tokenizer, token: *Token) {
+    token.type = .addition;
+
+    t += 1;
+    if t >= max_t then return;
+
+    current := t.*;
+    next    := (t + 1).*;
```
focus-editor, line 50: Looks like it can crash near the end of the file. Here and in the function below

SentientCoffee: Fixed by adding a `if t + 1 >= max_t` check on both.

src/langs/diff.jai
```diff
@@ -0,0 +1,147 @@
...
+
+    if t > max_t then t = max_t;
+}
+
+parse_minus :: (using tokenizer: *Diff_Tokenizer, token: *Token) {
+    token.type = .deletion;
+
+    t += 1;
+    if t     >= max_t then return;
+    if t + 1 >= max_t then return;
```
focus-editor, line 72: Only the second one of these ifs is required :)

PR #518, opened 2025-02-19
Add built-in function highlighting for Lua

SentientCoffee:
Built-in functions in Lua 5.4 now get highlighted as functions even without a succeeding (), because some people like to use `print "Something"` rather than `print("Something")` for example.

This was a feature request in the Discord from a few months ago. Sorry for the wait! I'd completely forgotten about it.


focus-editor:
Thanks for the PR.

Instead of doing a lookup this way we should just extend the keyword map -
see python.jai for an example

SentientCoffee:
Better this way. Fixed.

PR #522, opened 2025-02-26
Make '$' a valid JS identifier character

caztanj:


src/langs/js.jai
```diff
@@ -146,6 +147,32 @@ get_next_token :: (using tokenizer: *Js_Tokenizer) -> Token {
     return token;
 }
 
+read_utf8_identifier_string :: (using tokenizer: *Js_Tokenizer) -> string {
+    identifier: string;
+    identifier.data = t;
+
+    while t < max_t {
+        c := t.*;
+        if c == #char "_" || is_alpha(c) || (t > identifier.data && is_digit(c)) || c == #char "$" { t += 1; continue; }
```
focus-editor, line 156: Please add a compile-time parameter `identifier_chars: str` (or maybe you have a better name?) instead of copying the function manually.

caztanj: I moved the hard coded `_` to be the default value of the `identifier_chars` parameter. Do you want to have it this way or would you rather have the parameter be `extra_chars` and keep the `_` as it was before this change?

src/langs/common.jai
```diff
@@ -28,13 +28,13 @@ read_identifier_string :: (using tokenizer: *Tokenizer) -> string {
...
 
-read_utf8_identifier_string :: (using tokenizer: *Tokenizer) -> string {
+read_utf8_identifier_string :: (using tokenizer: *Tokenizer, $identifier_chars := "_") -> string {
     identifier: string;
     identifier.data = t;
 
     while t < max_t {
         c := t.*;
-        if c == #char "_" || is_alpha(c) || (t > identifier.data && is_digit(c)) { t += 1; continue; }
+        if contains(identifier_chars, c) || is_alpha(c) || (t > identifier.data && is_digit(c)) { t += 1; continue; }
```
focus-editor, line 37: `contains` is a bit heavy-handed for a very short predefined string that we have here. `array_find` is a better fit

src/langs/js.jai
focus-editor: Yes making it the default value is good

PR #531, opened 2025-03-15
Add theme keys for color preview title bar and cursor line highlight.

NotKyon:
Noticed that some elements I wanted to theme separately were tied together on `background4`, so I added separate keys for them.

`cursor_line_highlight` is named that way instead of `current_line_highlight` or similar due to multiple cursor support. (There isn't really a *current* or active line, per se.)

This change bumps the config version number from 16 to 17. The migration is intended to add the new keys using the existing `background4` key. The default value used for the new colors is the same default value as `background4`.

Wasn't able to test these changes myself, as Jai's in closed beta. It is possible the code doesn't compile or work as intended... however, because of how simple of a change this is, I am optimistic!


focus-editor:
We'll review and merge this after the editor-splitting branch is merged into main, because it touches the config migrator and will cause conflicts

PR #534, opened 2025-03-24
Improve touchpad scrolling on Wayland

caztanj:
This change makes touchpad scrolling scroll the distance the Wayland compositor tells us to scroll.

It also fixes a bug that made natural scrolling not work when it was enabled in the compositor.


ileonte:
Thank you for the fix!

Not sure I understand why this works but it does seem to work (at least it does on the compositors that I tested).

PR #542, opened 2025-04-23
C syntax highlighting: Highlighting for UL and ULL suffixes

CELLTH:
Color highlighting for UL and ULL suffixes. 

example:
```C
u64 foo = 1000000000ULL;
```


obiwanus:
Before dereferencing `t` we need to check that it hasn't run past the buffer's end

PR #543, opened 2025-04-29
Added setting for scroll_behind_first_line

kujukuju:
Very roughly added a setting for scroll_behind_first_line. I wrote this code like 6 months ago and just migrated it over to the main branch.

It doesn't include a setting for the number of lines, but I think it should...

### DIFF
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -1027,8 +1027,8 @@ draw_editor :: (editor_id: s64, main_area: Rect, status_bar_height: float, ui_id
         content_height := (num_editor_lines + 2) * line_height;
         scroll_behind_height := 0.0;
         scroll_behind_lines: s32 = 12;
-        if config.settings.scroll_behind_first_line && !as_build_panel && !is_color_preview_panel(editor_id) then content_height += (rect.h - 3 * line_height);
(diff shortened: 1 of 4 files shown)


src/draw.jai
```diff
@@ -1025,8 +1025,12 @@ draw_editor :: (editor_id: s64, main_area: Rect, status_bar_height: float, ui_id
         // Handle editor scrolling here because we only know the layout at this point.
         num_editor_lines := get_max_line_num(editor, buffer);
         content_height := (num_editor_lines + 2) * line_height;
-        if buffer_settings.scroll_beyond_last_line && !as_build_panel && !is_color_preview_panel(editor_id) then content_height += (rect.h - 3 * line_height);
+        scroll_behind_height := 0.0;
+        scroll_behind_lines: s32 = 12;
+        if config.settings.scroll_behind_first_line && !as_build_panel && !is_color_preview_panel(editor_id) then content_height += (rect.h - 3 * line_height);
+        if config.settings.scroll_beyond_last_line && !as_build_panel && !is_color_preview_panel(editor_id) then scroll_behind_height = (scroll_behind_lines * line_height);
```
focus-editor, line 1031: I think the settings here should be swapped

focus-editor: Also I just realised it should be using `buffer_settings` instead of `config.settings` because otherwise it won't be overridable for individual files (and it's marked as overridable in the settings struct)

kujukuju: fixed

PR #545, opened 2025-05-04
Fix crash when doing block-comment as fallback with multiple cursors

fbjerno:
Fixes a crash due to trying to calculate line start index from a dirty buffer after the previous cursor's selection was commented out. This change calculates the line start indices before any text is inserted.

I initially fixed this by running rescan_for_lines() for each cursor, but that seemed like a lot of unnecessary work, especially if working in a big file.

This bug was originally reported on Discord as follows:
> Focus crashes when you toggle comment (Ctrl-/) with multiple cursors in a .css file. 
Minimal reproduction: enter two empty lines, put a cursor on each and run the toggle-comment command. 
image

https://discord.com/channels/1106007785193345104/1354900494459473990

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -4580,25 +4580,39 @@ toggle_block_comment :: (editor: *Editor, buffer: *Buffer, is_fallback := false)
     } else {
         comment_total_length := cast(s32)(comment_start.count + comment_end.count);
 
-        for * cursor : editor.cursors {
-            move_count := cast(s32)it_index * comment_total_length;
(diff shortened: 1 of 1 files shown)


fbjerno:
I also noticed that un-commenting block comments can be unintuitive in some cases.
Namely, attempting to toggle comment (with fallback) when the cursor is in the middle of a multi-character start/end comment symbol will wrap the line in yet another block comment.

image

\*toggle_comment*

image

It would be nice to detect this and to remove the comment instead.

PR #550, opened 2025-05-25
Add highlighting for Vue files

caztanj:


src/langs/xml.jai
```diff
@@ -391,6 +392,12 @@ parse_html_attribute_name :: (using tokenizer: *Xml_Tokenizer, token: *Xml_Token
         return;
     }
 
+    if vue_mode {
+        attribute_value_is_js   = t.* == #char ":";
+        attribute_value_is_js ||= t.* == #char "@";
+        attribute_value_is_js ||= t + 1 < max_t && t.* == #char "v" && (t + 1).* == #char "-";
```
obiwanus, line 398: I would need to double check this, but I think it's not safe to dereference `t` when it equals `max_t`. Please can you confirm that this can't be used to crash the editor when this is the last token in the file.

PR #558, opened 2025-06-21
Small fix to Gcc error regex

victor-Lopez25:
Not sure what `(\[(?P<msg1>.*)\])` is for but if it isn't required, the space separating it from `(?P<msg>.*)` also shouldn't be required (as demonstrated from a couple examples):

All of these are not caught with the previous regex and are caught with this modification (only the first line of the error message is included):
`src/efi.h:374:5: error: unknown type name 'EFI_SIMPLE_INPUT_MODE'; did you mean 'EFI_SIMPLE_POINTER_MODE'?`
`src/efi.h:420:8: error: expected ';', ',' or ')' before 'UINT32'`
`src/efi.c:374:5: error: called object is not a function or function pointer`

There might be some example of `(\[(?P<msg1>.*)\])` being used but I don't remember any right now. I left it there in case it is used in some more obscure error messages.


focus-editor:
Please can you make a clean PR without conflicts (e.g. from a new branch) and I'll merge it. Thanks

PR #563, opened 2025-07-02
Add %FILE_LINE%, %FILE_COLUMN% to substitute_variables

fda0:
Querying file line and file column is useful for interacting with other programs like debuggers via commands.
For example: `run_command: raddbg --ipc find_code_location "%FILE%:%FILE_LINE%:%FILE_COLUMN%"`

### DIFF
--- a/src/build_system.jai
+++ b/src/build_system.jai
@@ -492,14 +492,23 @@ print_to_output_panel_from_main_thread :: (s: string, mark_as: Buffer_Region.Kin
 substitute_variables :: (using work: *Build_Thread_Work) {
     // Figure out the current file
     file := "";
+    coords_under_cursor: Coords;
     active_pane_editor_id := get_active_editor_id();
     if active_pane_editor_id > 0 {
         buffer := open_buffers[open_editors[active_pane_editor_id].buffer_id];
-        if buffer.has_file then file = copy_string(buffer.file.full_path);
+        if buffer.has_file
+        {
+            file = copy_string(buffer.file.full_path);
+            if buffer.cursors.count > 0
+            {
+                cursor := buffer.cursors[0];
+                coords_under_cursor = offset_to_real_coords(buffer, cursor.pos);
+            }
+        }
     }
 
     file_vars_are_used := false;
-    file_vars := string.["%FILE%", "%FILE_DIR%", "%FILE_NAME%", "%FILE_NAME_NO_EXTENSION%"];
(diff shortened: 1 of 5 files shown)


src/build_system.jai
```diff
@@ -497,14 +497,23 @@ print_to_output_panel_from_main_thread :: (s: string, mark_as: Buffer_Region.Kin
...
     file := "";
+    coords_under_cursor: Coords;
     active_pane_editor_id := get_active_editor_id();
     if active_pane_editor_id > 0 {
         buffer := open_buffers[open_editors[active_pane_editor_id].buffer_id];
-        if buffer.has_file then file = copy_string(buffer.file.full_path);
+        if buffer.has_file
+        {
+            file = copy_string(buffer.file.full_path);
+            if buffer.cursors.count > 0
```
focus-editor, line 507: I haven't had a chance to properly look at this PR, but a couple of things I've noticed after taking a quick look:

1. Cursors are owned by editor, not buffer - buffer cursors are only used for undo/redo. Please get the main cursor of the editor (not the one with the index 0) for this.
2. Please follow the style of the surrounding code. We never put an opening brace on a new line.

Thanks!

src/build_system.jai
```diff
@@ -516,26 +525,36 @@ substitute_variables :: (using work: *Build_Thread_Work) {
 
         file_dir, file_name_no_extension, _, file_name := path_decomp(file);
         file_dir = trim_right(file_dir, "/");
+        file_line := sprint("%", coords_under_cursor.line + 1);
+        file_column := sprint("%", coords_under_cursor.col + 1);
```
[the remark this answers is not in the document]
fda0: I don't exactly know what the codebase conventions are as I haven't seen 99% of it. Let me know if you feel that this should be changed.

My reasoning here was that other string-allocating functions in `substitute_variables` are using default allocator.
We call `copy_string(buffer.file.full_path)` and not `copy_string(buffer.file.full_path,, temp)` or `copy_temporary_string(buffer.file.full_path)`.
The convention seemed to be that one calls `push_allocator` outside of `substitute_variables`.

`sprint` uses default allocator while `talloc` uses temporary allocator. Seemed more consistent to use default allocator.

src/build_system.jai
focus-editor: Your original names are good, even if there's no underlying file people should still find it intuitive

Issue #511, opened 2025-02-04
add .tpp and .ipp extensions recognizeation for c++


rkc137:
ipp and tpp is something like .cpp but for templates code
[what is tpp/ipp exactly](https://cplusplus.com/forum/beginner/272582/)

focus-editor:
Will be in the next release. In the meantime you could use file associations to assign syntax to those types of files, like this

```
[[workspace]]
...
[file associations]
*.tpp *.ipp : cpp
```

PR #565, opened 2025-07-13
Use C++ syntax highlighting for .tpp files

fbjerno:
This change applies C++ syntax highlighting to .tpp files by default.
Fixes https://github.com/focus-editor/focus/issues/511


src/editors.jai
```diff
@@ -2197,6 +2197,7 @@ get_lang_from_path :: (full_path: string) -> Buffer.Lang {
         case "hh";              lang = .Cpp;
         case "m";               lang = .Cpp;
         case "mm";              lang = .Cpp;
+        case "tpp";             lang = .Cpp;
```
focus-editor, line 2200: Shouldn't this also include `ipp` to fix #511 ?

PR #572, opened 2025-07-30
Fix unsaved files overflow

fbjerno:
This change modifies `draw_generic_confirm_dialog()`, so it actually passes all the available extra space to the extra draw function. Previously it only passed a rectangle the size of the dialog message. `draw_unsaved_buffers_dialog()` now uses `extra_rect` to calculate how many unsaved files it has space to display. If it doesn't have space for all unsaved files, it will also display a message saying how many files were not displayed.


focus-editor:
Looking very nice

PR #578, opened 2025-08-12
Add option to show/hide tab characters

fbjerno:
Adds a setting to show/hide tab characters. 

A global in the Simp module controls which glyph is used for tab characters; an arrow when enabled, a space (nothing) when disabled. When the setting is changed, the glyph currently used for tab characters is removed from the glyph lookup of all fonts, forcing it to be recreated the next time it is used.

Currently I think the glyph ends up being redrawn by FreeType every time the setting is changed, which isn't ideal - the desired glyph may well already have been rendered. But hey, how often are people really going to be enabling/disabling tabs?

### DIFF
--- a/config/themes/basic-light.focus-theme
+++ b/config/themes/basic-light.focus-theme
@@ -1,4 +1,4 @@
-[23]  # Version number. Do not delete.
+[24]  # Version number. Do not delete.
 
 [colors]
 background0:                            FFFFFFFF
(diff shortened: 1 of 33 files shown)


focus-editor:
Thanks, I will review this when I can - busy now

focus-editor (commented):
Nice, this looks much better.

I see you've bumped the config version but haven't updated the unit tests - they will fail when trying to do a release build

fbjerno:
Aha, gotcha. Updated the unit tests now. Will try to remember to do a release build before pushing in the future.

PR #579, opened 2025-08-13
Add Powershell syntax highlighting

JeremyLullaby:
First experience with jai, and no compiler access to test, but I believe this addition covers most of the tokens utilized by Powershell. Hope this helps.


focus-editor:
It almost works, with things from @fda0 and a couple of fixes on top. It needs some more work however. I will see if I can add it 

image

Issue #393, opened 2024-08-05
[Feature Request] Revamp project switching


maxamundsen:
(Similar issue to #360) 

In my opinion, the user experience of loading project files can be significantly improved. Currently, if you want to load a project outside the global project dir (perhaps you keep the project in VCS), you need to exec focus from the CLI in a dir containing a `.focus-config`. This is cumbersome, and it would be nice to be able to load these files directly from focus without needing the terminal. It would also be nice to keep a list of recently loaded project files that may be located across the system for quick access. This change would remove the need for a global project directory since projects can be referenced "globally" via the filesystem, or quickly via the "recents" menu.

With these changes, it would also be very handy to reference files/directories inside a project _relative to the project file_, instead of needing to specify an absolute path. This would apply to `[[workspace]]`, `[ignore]`, `[allow]`, and `[[build commands]]`.

**Modified commands:**
- "Switch To Project":  Instead of searching the global projects folder, prompt user to find a file with a `.focus-config` extension. If the file is a valid project file, load it, and push the name, and path to a `recent-projects.txt`. If that project was already in the recents list, move it to the top of the list.

**New commands:**
- "Switch to recent project": Displays a list of recent project names (and perhaps their path) specified in `recent-projects.txt`. Loading a project will always move it to the first position in the list, displaying it first in subsequent calls to the "switch to recent project" command.
- "Clear recent projects": Clears list of recent projects.

**New options for global config:**
- `open_recent_project_on_launch` - Boolean that would determine if focus should load the most recent project in the recent list on program init.

**Changes to project config:**
- Add a `%PROJECT_CWD%` variable, which would resolve to the directory of the project file. This would allow for configuration relative to the project file for build commands, and ignore/allow dirs.

Obviously any commands that utilize the 'recent file' will attempt to access/modify a file on disk, so an error message would need to display if the file is inaccessible.

focus-editor:
Looks like these things are mostly implemented?

maxamundsen:
Yes! The only thing I would add to this is the ability to load a project from a directory containing a `.focus-config` via the editor itself, and not the CLI.

Somebody in the discord suggested the following: Add a dialog that allows you to open a directory. If that directory contains a `.focus-config`, load that project. If the selected directory is NOT a project, set that directory to the current workspace as if you launched focus via the cli with that directory as an argument.

Otherwise these changes have been really nice and I'm glad that they were added. Thank you.

PR #595, opened 2025-09-16
Switch to project or create new

### LINKED ISSUE #152: Feature Request: Command to Initiate Project Configuration Template
**Description:**
While the "Open Config For Project" command is incredibly useful for viewing and adjusting project configurations, it's often the initial setup of these configurations that can be daunting for many users. To simplify this process, I propose the addition of a "Create Config For Project" command.

**Feature:**
Create Config For Project Command: When run, this command should generate and load a default project configuration template. This template would serve as a beginner-friendly starting point, assisting users in setting up their project's basic parameters without having to start from scratch or copy it manually from the global config.

The template could look like as follows:
```
[1]  # Version number. Do not delete.

[[workspace]]

[workspace dirs]
# These directories will be scanned when a workspace is opened so that search etc. works.
# Example:
# C:/projects/my-project
# C:/jai


# IMPORTANT:
# You can override any section from the global config in your project and it will be used when your project is active.
# Please note that some of the sections will completely replace the ones in the global config, while other sections
# will try to merge with the global config. E.g. the workspace sections such as [ignore dirs] will completely
# override the global [ignore dirs], which allows you to "unignore" some dirs just for your project!
# Empty sections are ignored.

[ignore dirs]
# The workspace scanner will not descend into the dirs that match these names.
# Example:
#.svn
#.git
#__pycache__

[allow file extensions]
# These files are explicitly allowed so that the editor doesn't waste time figuring out
# whether it should ignore these or not (if a file is not explicitly allowed or ignored,
# it will be read to determine if it's binary or not).
# Example:
# .jai .c .cpp .h .hpp

[ignore file extensions]
# Files with the following extensions will not appear in the file open dialog and won't be loaded into memory.
# It makes sense to add all known binary file extensions here.
# Example:
# .exe .pdb .ttf .otf .eot
```

filippocrocchini:
This closes #152.
When switching to a new project allow for creating a new one the same way as you do by saving a file. 
This also include a small fix to the open file dialog, it no longer shows the save as option if the file name matches exactly with an existing file.

This is what it looks like:
image

After a new project is created and switched to, it will open the newly created project config so that you can edit it.

I added "Open or Create a Project" to the welcome screen. I also tweaked the vertical alignment of the right side so that it's visually centered with the left side.
image

### DIFF
--- a/src/deferred_actions.jai
+++ b/src/deferred_actions.jai
@@ -10,8 +10,9 @@ execute_deferred_actions :: () {
     }
 }
 
-defer_action_switch_to_project :: (path: string) {
+defer_action_switch_to_project :: (path: string, open_project_file := false) {
     clear_action_queue();
+    defer_action_open_file(path);
     push_action(.reload_workspace, .{ reload_workspace = .{ automatic = false } });
     push_action(.load_project_config, .{ load_project_config = .{ path = copy_string(path) } });
     push_action(.abort_workspace_scanning);
@@ -112,6 +113,9 @@ defer_action_save_buffers :: (buffer_ids: [] s64) {
     push_action(.save_buffers, .{ save_buffers = .{ buffer_ids = array_copy(buffer_ids) } });
 }
 
+defer_action_open_file :: (path: string, placement: Editor_Placement = .in_place ) {
+    push_action(.open_file, .{ open_file = .{ path = copy_string(path), placement = placement } });
+}
 
 Deferred_Action_Id :: #type,distinct s64;
 
@@ -150,6 +154,7 @@ execute_top_action :: () -> Action_Result {
         case .close_all_editors;                        return execute_close_all_editors(action);
         case .close_pane;                               return execute_close_pane(action);
         case .close_other_panes;                        return execute_close_other_panes(action);
+        case .open_file;                                return execute_open_file(action);
     }
 }
 
@@ -173,6 +178,10 @@ remove_top_action :: () {
             using action.details.save_buffers;
             array_free(buffer_ids);
 
+        case .open_file;
+            using action.details.open_file;
+            if path { free(path); path = ""; }
+            
         // Nothing to free
         case .quit;
         case .save_as;
@@ -216,6 +225,7 @@ Deferred_Action :: struct {
         close_editor:               struct { editor_id: s64; }
         close_editors_for_pane:     struct { pane: *Editor_Pane; except: s64; notify: bool; }
         close_pane:                 struct { pane: *Editor_Pane; }
+        open_file:                  struct { path: string; placement: Editor_Placement; }
     }
 
     Kind :: enum {
@@ -238,6 +248,7 @@ Deferred_Action :: struct {
         close_all_editors;
         close_pane;
         close_other_panes;
+        open_file;
     }
 }
 
@@ -469,7 +480,7 @@ execute_refresh_config_or_theme :: (action: Deferred_Action) -> Action_Result {
 
 execute_reload_workspace :: (action: Deferred_Action) -> Action_Result {
     reload_workspace(action.details.reload_workspace.automatic);
-    return .CANCEL;  // final command
+    return .DONE;
 }
 
 execute_build_command :: (action: Deferred_Action) -> Action_Result {
@@ -540,6 +551,13 @@ execute_close_pane :: (action: Deferred_Action) -> Action_Result {
 
 execute_close_other_panes :: (action: Deferred_Action) -> Action_Result {
     close_other_panes();
+    return .DONE;
+}
+
+execute_open_file :: (action: Deferred_Action) -> Action_Result {
+    print("Before opening file.\n");
+    success := editors_open_file(action.details.open_file.path, action.details.open_file.placement);
+    print("OPENING FILE % -- %\n", action.details.open_file.path, success);
     return .DONE;
 }
 
--- a/src/draw.jai
+++ b/src/draw.jai
@@ -182,6 +182,7 @@ draw_welcome_screen :: (main_area: Rect) {
     margin      := logo_rect.x;
     shrink_step := logo_rect.w / 6.5;
 
+    left_y_cursor : float;
     // Draw the left part
     {
         draw_rect(left_area, .BACKGROUND_3, extra_draw_info = .{is_background = true});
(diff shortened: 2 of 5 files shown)


filippocrocchini:
Let me know what else needs to be changed, I'm still figuring out the codebase.

focus-editor:
Thanks for your MR - I will review this when I can, quite busy now, sorry

focus-editor:
I'm still busy but I've had a quick look and I can already say that I don't want to mix switching to project and the creation of a new project. 

Instead the plan was to add a new dialog where the user would just type the name of the project without specifying any path and it would save it to the global project directory. Then we wouldn't have to pollute the existing dialogs with this extra logic and also would avoid the users accidentally creating new projects when they mistype the name of the project they want to switch to.

Also, the convention going forward is to keep the name of the action as close to the display name of the action as possible - i.e. we shouldn't call an action "Open Project" when the underlying action in code is called `switch_to_project`

filippocrocchini:
I see, yeah I was unsure of how you wanted to approach the issue. My idea was to try to contain the amount of commands. In this case I agree that having a dedicated "new project" probably makes more sense. I'll see if I can add that instead in the coming days. Thank you for the feedback!

filippocrocchini:
I have some questions I have been thinking about:

1. How would you prefer handling creating a project with a name that already exists? Maybe doing something similar to how github handles naming a repository? E.g. showing "Project name is available." / "Project name is already taken." under the input box.
2. Do we want to allow people to create a project in the a specific directory instead of the global one? And in which case, do we want to use the save file dialog box, with the default path to the global config folder? That way we solve issue 1 for free as well.
3. Should new_project also switch to it once it's created, and open the config? Instead of having to do `new_project` -> `switch_to_project` -> `open_project_config`.
4. What would you think the new project-config should contain:
    1. Default/example config
    2. Global config copy
    3. Default/example config but modified --ideally commented out-- to avoid overriding global config keybindings/colors/setting and whatnot

focus-editor:
> What would you think the new project-config should contain

Some minimal contents, maybe similar to what the Example Project file contains

filippocrocchini:
Nice, I would love to work on this! I'm ok with doing some back and forth. It might take a while but that's ok. I'll check #393 and also what you are doing with config and file watcher so I know what not to touch. This is a bigger project than I was expecting 😅 but I think it will improve usability a lot. I think I will get back to you with a plan before I actually start changing stuff around.

filippocrocchini:
So, initially I was thinking of adding a bunch of commands like "open project file", "new project", "new project file", "new project from folder" and so on but it didn't feel like it fit focus. There were too many edge cases. So here's what I think is best, which would build on the system as it currently is. Let me know what you think. If you like the idea I will start working on a proof of concept implementation.

We have the concept of project config which only exists when a project is open. My plan is to extend this concept to always existing even if no project is open. Think of it as a "workspace" config. To do so, when no project is open there will still be a config which will act as backed by a file, but it won't be (I guess it could also be backed by a temporary file). Meaning, you can save it, which will reload the "project" with the changes you made. And save it as, which will effectively create a new project by making it backed by a file. Once this core feature is there we can get everything else we want organically. I believe. 

## New/modified commands
- Remove show_default_config, this should no longer be necessary.
- Add switch_to_project_file.
- Add add_folder_to_project.
- Modify switch_to_project to display recent projects as well.
- Maybe add save_workspace_as_project, but I don't think it's strictly necessary.

Note:
- open_project_config will open the virtual config if no project is loaded.
- close_project will reset the virtual config if no project is loaded, prompting for saving as it's a destructive operation.
- switch_to_project_file: Pick a focus-config from file picker.   
- switch_to_project: Pick from global *and recent* projects.
    - Question: Is there a reason why we can't use the file picker opened in the global config folder here? That way we don't need the "Switch to project file" command.

## Examples
All of the following examples assume you have no project open, but will work if you do have one open. They will just achieve something different.

- Create a new project:
    1. open_project_config (virtual config)
    2. save_as (alternatively we can have a save_workspace_as_project command)
- Switch to a project that is not in the global config folder:
    1. switch_to_project_file
- Open a folder 1:
    1. open_project_config (virtual config)
    2. Add folder to [Workspace]
    3. save (or save_as to create a new project)
- Open a folder 2:
    1. add_folder_to_project (this way you can use file picker if you need to browse)

focus-editor:
>  "open project file", "new project", "new project file", "new project from folder" and so on but it didn't feel like it fit focus

Not sure why those commands would be needed, you're right that we should keep things simple.

> My plan is to extend this concept to always existing even if no project is open

I feel like this is already overcomplicating things. If no project is active, the user shouldn't be able to open its config file. Also remember that changing the way configs get loaded and applied would conflict with my config system rewrite.

> Remove show_default_config, this should no longer be necessary.

I'm not seeing why it would no longer be necessary - it's still useful for people to reference after their custom configs have diverged from the default one.

> Add switch_to_project_file.

Why? There's already `switch_to_project`, no need for another command - if the user wants to load a project from a file, there should be an option in the project list itself, which would then open a file picker.

> Add add_folder_to_project.

I would suggest to call it `add_directory_to_workspace` - we don't use the word folder anywhere in the actions, and `workspace` always exists, even if there's no project. 
Ideally the same thing should happen when a user drags a folder into the editor (but that could be something for later).

> Modify switch_to_project to display recent projects as well.

I think it already acts like this.

> Maybe add save_workspace_as_project, but I don't think it's strictly necessary.

It is useful and it was in the plan. It should also be easy to implement - just make it do what the `create_new_project` command would do, but when creating the project file just populate the workspace with whatever the user has loaded.

To reiterate, I don't think we should be introducing virtual projects. Everything that was planned as part of this work should be naturally implementable without adding new concepts

filippocrocchini:
> I feel like this is already overcomplicating things. If no project is active, the user shouldn't be able to open its config file. Also remember that changing the way configs get loaded and applied would conflict with my config system rewrite.

Sure, that's a good point. If for example you want to open a folder **without creating a project first** as you can do in many other editors, be it in memory or in a temp file, somewhere there must be a temporary "workspace config" that contains the currently tracked folder/s and so on. So, my proposal was to make that accessible to the user until it becomes a project file, that way they can change settings and add folders the same way they normally would. This would be the same behavior as VSCode. There is a workspace settings tab if you have a folder open (or a project obviously), they will create .vscode folder with workspace settings when you commit changes; if you have multiple folders in the workspace they will create a temporary workspace in AppData until you save it. I do concede that the naming it project-config might be confusing there, so following them and calling it workspace-config would be best imo.

I think in general it's a common pattern to open a text editor, open some folders, change some configs and then once it's set up, save the workspace to a file as a project --or maybe not--. So, I wanted to facilitate that workflow by delaying asking the user to make choices until it's unavoidable (e.g. picking where the project should go). Moreover it would remove the need for a new_project command, as closing the current project and adding a folder to the workspace (or opening a new focus instance) would be enough.

If you think there could be conflict between implementing this and your config and file watcher rewrite let me know and we can pause this until those are done. If instead you prefer not allowing users access to settings until they have created a project I understand and will move on with that.

focus-editor:
> somewhere there must be a temporary "workspace config" that contains the
currently tracked folder/s and so on

There's the in-memory config which contains everything, when the user opens
a folder, that in-memory workspace dir array will get modified, then if the
user wants to save that as a project, the information can be taken from
there to generate a config file. No need to invent any new concepts.

focus-editor:
I totally understand. Let's keep it open for now and then we'll see if we're able to continue this. Thanks for going this far with this.

PR #598, opened 2025-10-02
Enable ARM64 build on macOS

caztanj:


focus-editor:
That's great, did you adapt File_Async to support arm64? Did you send it to be included in modules by default?

The only concern I have is how performant the fallback hash will be, as I imagine it's going to be much slower than meow_hash - need to check whether the arm64 version is slower than the x64 one on large files

focus-editor:
Also, the release build script needs to be updated to not create an `-Intel.dmg` file

caztanj:
> That's great, did you adapt File_Async to support arm64? Did you send it to be included in modules by default?

Yes, I sent the changes I made to `thread_pool.jai`.

PR #607, opened 2025-10-17
Buffer error highlighting

sborkowski:
This adds simple but fairly robust error highlighting to buffers.

imageimage

It's pretty simple:
- On build, error regions are interpreted and added to each corresponding buffer.
  - Error regions are separate from other regions, because the 'region' array seems to get nuked sometimes.
- The region contains the entire text part of the line (trimmed).
  - This is because the column numbers are unreliable between compilers and because no language parsing is happening.
  - This also means that anything like a comment at the end of the line will get included.
- There is logic to update the regions when the buffer changes.
  - The logic is such that the region follows text edits. You can grow and shrink it, you can crop it, and you can delete it, but you can't bring it back with undo actions. A rebuild would be required to see it again. Seems like a reasonable trade-off though.
  - Currently this is only suited for errors, but it could be adapted for other types of regions.
- Buffer is highlighted and scrollbar is marked with the use of these regions inside the draw loop.

Overall, the interaction feels pretty good to me.

Future considerations:
- Add a setting to highlight the entire line, perhaps with a lower opacity, for better visibility.
- Add other region highlights like, note and to-do comments. The current update logic is specifically for errors, so some adjustments would need to be made. For example, if you edit any part of "TODO", then the highlight should be removed. You probably want some sort of pooling for this reason, as Array doesn't handle removal.
 
Anyways, I'm still wrapping my head around the code base, so that will be for another time.

### DIFF
--- a/src/editors.jai
+++ b/src/editors.jai
@@ -538,18 +538,21 @@ refresh_all_editors_for_buffer :: (buffer_id: s64) {
                 }
             }
 
-            for *region : buffer.errors {
-                // Adjust or remove regions
-                if edit.start <= region.start && edit.end <= region.start {
-                    region.start = region.start + edit.delta;
-                    region.end = region.end + edit.delta;
-                }
-                else if (edit.start > region.start && edit.end < region.end  ) ||
-                        (edit.start < region.start && edit.end > region.start) ||
-                        (edit.start < region.end   && edit.end > region.start) ||
-                        (edit.start < region.end   && edit.end > region.end  ) {
-                    region.start = 0;
-                    region.end = 0;
+            if edit.editor_id != editor_id // only if we haven't updated this buffer before.
+            {
+                for *region : buffer.errors {
+                    // Adjust or remove regions
+                    if edit.start <= region.start && edit.end <= region.start {
+                        region.start = region.start + edit.delta;
+                        region.end = region.end + edit.delta;
+                    }
+                    else if (edit.start > region.start && edit.end < region.end  ) ||
+                            (edit.start < region.start && edit.end > region.start) ||
+                            (edit.start < region.end   && edit.end > region.start) ||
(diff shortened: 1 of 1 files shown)


focus-editor:
Thanks for submitting the PR.

I've tried it and found that while the region adjustment works well when edits are made above, below or inside the region, it doesn't work as well when the region boundary is affected by an edit (e.g. when deleting a range which includes one of the region boundaries). It will then break and there will be no way to dismiss it, which is annoying. 

I think error regions should just be removed on any edits. I don't remember that part of the code well enough now but I think if we just use `buffer.regions` then this behaviour should happen naturally.

sborkowski:
I see what you're saying. I do kind of want to fix the region updating, because I think I can get it to work 'properly'. There will still be some weirdness, like if you get two regions next to one another, then you lose track of the ends as you edit. Maybe that's an acceptable consequence of this? I don't know. Anyways, if you feel strongly that they should just disappear on edit, I can make that happen instead. That was actually my first implementation anyways.

focus-editor:
If you want you can try fixing it, but if it doesn't feel good we might have to throw it away anyway - would you be okay with that?

I suggest that you have a look at this piece of code which adjusts cursors on edits. It works well and should be easily adaptable for region adjustment.
image

Also, if we do keep the error regions even after edits they will need to be dismissable on escape - the `escape` action needs to be updated to remove them similarly how other actions are handled there

sborkowski:
The range is needed because you can span multiple error regions, and if you insert text, the delta only tells you the final change. For example, if let's say you select 100 characters, and paste 50. The delta is 50. If the offset is somewhere behind the error region, how can you know if the change overlaps the region? You can't without providing the range.

In any case, I understand your hesitation and I don't want to waste your time. Feel free to close this, but I'll probably keep cracking away at it, as I'd like to have the feature even if its for myself.

PR #617, opened 2025-11-19
Changes for error/warning-less compilation with 0.2.019

rmtew:
Required the following:
* Replace `for #v2` with `for` as it is implicitly the case now (not backwards compatible with pre-0.2.019 I suspect).
* Remove automatic dereferencing of pointers passed into non-pointer functions (backwards compatible I expect).


obiwanus:
Thanks!

PR #620, opened 2025-11-28
Correct the behavior of x11_wait_for_events

caztanj:
This change fixes the very annoying input lag on the X11 backend.

There is no need to call XQLength or XPending in this procedure and in case we have events on multiple fds we need to handle all of them in the current iteration instead of breaking on the first.


caztanj:
That is not something I have noticed so I don't know.

PR #640, opened 2026-02-21
Add support for drag and dropping directories

caztanj:


caztanj:
Now it re-scans the workspace.

PR #648, opened 2026-04-03
Add ability to sort lines

hovsater:
**Disclaimer:** This is the first time I've written anything in Jai and it's also my first time exploring the Focus codebase. Bare with me.

I'm still not sure how we should handle trailing lines at the end of a range (e.g., lines only consisting of ` `, `\t` and `\n`). Right now I strip any `\n` at the end of the range. It kind of works, but is perhaps somewhat dirty?


hovsater:
Fixed in https://github.com/focus-editor/focus/pull/648/commits/bdd66a08529ea6dcc678cb8a7f8100102f719e30.

PR #651, opened 2026-04-10
Fix allocator rename for jai 0.2.028

pikkle:


src/main.jai
```diff
@@ -1111,7 +1111,7 @@ focus_allocator: Allocator;
     #import "Machine_X64";
 }
 
-using Basic :: #import "Basic"()(MEMORY_DEBUGGER = DEBUG, TEMP_ALLOCATOR_POISON_FREED_MEMORY = false);
+using Basic :: #import "Basic"()(MEMORY_DEBUGGER = DEBUG, TEMP_ALLOCATOR_OVERWRITE_MEMORY = false);
```
obiwanus: We should just remove the parameter since `false` is the default anyway. Then older compiler versions should still work.

PR #656, opened 2026-05-11
Update to Jai 0.2.029

caztanj:
I changed the Details union to a tagged union since struct literals can no longer be used to assign values to untagged unions.

### DIFF
--- a/modules/File_Async/thread_pool.jai
+++ b/modules/File_Async/thread_pool.jai
@@ -95,7 +95,9 @@ initialize_queue :: ($User_Data: Type) -> *Queue(User_Data), result := Error.{} 
         td.name = sprint("Worker thread % for queue %\0", it_index, queue);
 
         ts := *td.ts;
-        set_initial_data(ts, TEMPORARY_STORAGE_SIZE, td.ts_data.data);
+
+        ts.data = td.ts_data.data;
+        ts.size = td.ts_data.count;
 
         td.thread_context = context;
         td.thread_context.temporary_storage = ts;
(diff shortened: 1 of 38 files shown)


obiwanus:
Looking good, but I haven't tested it so please let me know when you think it's ready to be merged 👍

caztanj:
I think it's ready now. I've used it a couple of days on macOS now and it also compiles and runs on both Windows and Linux now.

modules/File_Async/thread_pool.jai
```diff
@@ -95,7 +95,6 @@ initialize_queue :: ($User_Data: Type) -> *Queue(User_Data), result := Error.{}
         td.name = sprint("Worker thread % for queue %\0", it_index, queue);
 
         ts := *td.ts;
-        set_initial_data(ts, TEMPORARY_STORAGE_SIZE, td.ts_data.data);
```
obiwanus, line 98: Do we not need to set it manually?

modules/Linux_Display/linux_display.jai
```diff
@@ -43,7 +43,7 @@ init_display :: (display: *Display, desired_backend := BACKEND_CHOICE.PREFER_BES
 
     if display.base.type != Base_Display return true;
 
-    display.* = .{};
```
obiwanus, line 46: Does this assignment no longer work?

modules/File_Async/thread_pool.jai
caztanj: Yes, apparently i forgot to check it in.

modules/Linux_Display/linux_display.jai
caztanj: No, it gives this error message:

```
focus/modules/Linux_Display/linux_display.jai:46,17: Error: Struct literals cannot be used to provide values for untagged unions.

    if display.base.type != Base_Display return true;
    display.* = .{};
                ^^^

focus/modules/Linux_Display/linux_display.jai:8,12: Info: Here is the type of the expression.

    }
    Display :: union {
```

PR #661, opened 2026-06-03
Increase MAX_FONT_SIZE limit

rexim:
Also fix the bug discovered on the stream. Please let me know if I should change anything. Thank you!


obiwanus:
Noice!

obiwanus:
image

obiwanus:
Also it might've been me who introduced the bug while porting font.jai, not Jon, since it seems they have a list of pointers in there. Just to clear Jon's name haha

PR #675, opened 2026-07-26
Windows: Use native shell APIs in the correct way to support third-party file managers

yasinkaraaslan:
My previous PR had a bug, I didn't test it against highlighting files. Sorry about that again. This should work with both Windows Explorer and Filepilot hopefully. 

If you do Ctrl-Shift-P + Open File Directory or navigate to a folder and press Shift-Enter, it calls to ShellExecuteW for that folder (this was the part that worked). If you press Shift-Enter while highlighting a file, it calls to SHOpenFolderAndSelectItems and if that fails, it fallbacks to ShellExecuteW and opens the folder the file is in. (Which seems to be [how Chrome does it](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/chrome/browser/platform_util_win.cc) as well.)

So both paths work correctly for Windows Explorer, but Filepilot falls back to opening the folder the latter case since it fails on SHOpenFolderAndSelectItems. (maybe because it doesn't register the COM stuff it's supposed to, I don't know)

### DIFF
--- a/src/platform/windows.jai
+++ b/src/platform/windows.jai
@@ -215,11 +215,24 @@ platform_get_fonts_dir :: () -> string {
 }
 
 platform_open_in_explorer :: (path: string, reveal := false) {
-    path_backslashes := copy_string(path,, allocator = temp);
-    path_overwrite_separators(path_backslashes, #char "\\");
+    ITEMIDLIST :: struct {}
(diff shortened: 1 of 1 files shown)


yasinkaraaslan:
I closed it because it seems to stall for Roman-Skabin when using Filepilot.

